PR #1, opened 2022-02-24
Add GitHub Actions for CI

kthompson23:
This PR adds a CI GitHub Action to lint, typecheck, format, and run the test suite on all pushes.

I grouped the lint, typecheck, and format scripts into one job and the test in a separate job to run in parallel with the lint/typecheck/format job. For the test job I used the matrix option to run tests against node 12, 14, and 16.

A new test script `test:ci` was added to include the options we use in our service test suites.


shawnmcknight (approved):
I don't know the format of the actions, but the steps look right to me.

PR #14, opened 2022-03-01
Convert Schema, Document, and schemaType from JS to TS

shawnmcknight:
This PR converts the `Schema` class, the `Document` class, and the various `schemaType` classes from JavaScript to TypeScript.  While working on this conversion, several changes were made to the codebase for one or more of the following reasons:
* The TypeScript compiler alerted to errors that were previous undetected and were fixed as part of the change
* The TypeScript compiler alerted to codepaths that were not really accessible without violating type rules and the codepaths were removed or altered accordingly
* The TypeScript compiler alerted to inconsistencies in API structures in subclasses

I will attempt to note as many of the changes as I can inline at the point of the changes.

During the conversion, all existing tests were scrapped and rewritten from scratch.  I believe the tests are much more solid now and exercise the library much better than before.

Additionally, some cleanup was performed to ensure all tests are in `__tests__` folders rather than as siblings of the unit under test.

This PR is targeting a topic branch `ts-conversion` so the typescript changes can be staged in that branch.


jest.config.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
 	testEnvironment: 'node',
 	resetMocks: true, // clear history and reset behavior of mocks between each test
 	restoreMocks: true, // restore initial behavior of mocked functions
+	testPathIgnorePatterns: ['/node_modules/', '/dist/'], // ignore dist folder
```
shawnmcknight, line 5: I kept getting alerts from jest about collisions in the hastemap, so I added this to stop them.

src/ts/Document.ts
```diff
@@ -0,0 +1,223 @@
...
+	/** Schema instance which defined this document */
+	private readonly schema: Schema | null;
+
+	/** Record array of multivalue data */
+	private record: MvRecord;
+
+	/** Indicates whether this document is a subdocument of a composing parent */
+	private readonly isSubdocument: boolean;
+
+	public constructor(schema: Schema, options: DocumentConstructorOptions);
```
shawnmcknight, line 55: I've added overloads here to try to ensure that the constructor is called appropriately.

src/ts/Document.ts
```diff
@@ -0,0 +1,223 @@
...
+	public constructor(schema: null, options: DocumentConstructorOptionsRaw);
+	public constructor(schema: Schema | null, options: DocumentConstructorOptions) {
+		const { data, isSubdocument, record } = this.parseConstructorOptions(options);
+
+		this.schema = schema;
+		this.record = [];
+		this.isSubdocument = isSubdocument;
+		this.transformationErrors = [];
+
+		Object.defineProperties(this, {
```
shawnmcknight, line 65: This was changed to remove the class methods from the definitions.  All functions were changed from properties to class methods which cannot and do not need to be adjusted by the Property Descriptors (because they are not properties).  They are not enumerable so changing the definition isn't necessary anyway.

src/ts/Document.ts
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	/** Transform document structure to multivalue array structure */
+	public transformDocumentToRecord(): MvRecord {
+		return this.schema === null
+			? getIn(this, '_raw', [])
+			: Object.entries(this.schema.paths).reduce(
+					(record, [keyPath, schemaType]) => {
+						const value = getIn(this, keyPath, null);
+						return schemaType.set(record, schemaType.cast(value));
```
shawnmcknight, line 89: I added a call to `schemaType.cast` here of the value.  The previous behavior relied on `validate` having been called before `transformDocumentToRecord` which is not a good assumption.  Adding the `cast` here allows it to be called in any order.

src/ts/Document.ts
```diff
@@ -0,0 +1,223 @@
...
+		if (this.schema === null) {
+			return [];
+		}
+
+		// U2 does not allow commas in filenames so we can use it to separate filename/entityName combinations
+		const separator = ',';
+		const definitionMap = Object.entries(this.schema.paths).reduce(
+			(foreignKeyDefinitions, [keyPath, schemaType]) => {
+				const value = getIn(this, keyPath, null);
+				const definitions = schemaType.transformForeignKeyDefinitionsToDb(schemaType.cast(value));
```
shawnmcknight, line 106: Similar note here regarding `schemaType.cast()`.

src/ts/Document.ts
```diff
@@ -0,0 +1,223 @@
...
+				// Deduplicate foreign key definitions by using a filename / entity name combination
+				// We could deduplicate using just the filename but ignoring the entity name could result in confusing error messages
+				definitions.forEach(({ filename, entityId, entityName }) => {
+					const key = `${filename}${separator}${entityName}`;
+					const accumulatedEntityIds = foreignKeyDefinitions.get(key) ?? new Set<string>();
```
shawnmcknight, line 111: There was a comment about switching to nullish coalescing here for node 14, but Babel will transpile this fine for Node 12 support.

src/ts/Document.ts
```diff
@@ -0,0 +1,223 @@
...
+					// For array types we may need to validate multiple foreign keys
+					accumulatedEntityIds.add(entityId);
```
shawnmcknight, line 113: There was an issue here where `Set.add` was being called with a spread against an array.  That is not valid syntax so I addressed it here.

src/ts/shared/classes/ForeignKeyDbTransformer.ts
```diff
@@ -14,22 +14,22 @@ export type CompoundForeignKeyDefinition = PositionForeignKeyDefinition & {
...
-		foreignKeyDefinition: ForeignKeyDefinition | CompoundForeignKeyDefinition | null,
+		foreignKeyDefinition: ForeignKeyDefinition | CompoundForeignKeyDefinition | null = null,
 	) {
 		this.foreignKeyDefinition = foreignKeyDefinition;
 	}
 
 	/** Transform schema foreign key definitions to the format required by the db server */
-	public transform = (value: string | null): ForeignKeyDbDefinition[] => {
-		if (value == null || this.foreignKeyDefinition == null) {
+	public transform = (value: unknown | null): ForeignKeyDbDefinition[] => {
```
shawnmcknight, line 31: Since any value could possibly be passed to this method, I had to loosen the type of the parameter.

src/ts/shared/classes/ForeignKeyDbTransformer.ts
```diff
@@ -14,22 +14,22 @@ export type CompoundForeignKeyDefinition = PositionForeignKeyDefinition & {
 
 export interface ForeignKeyDbDefinition {
 	filename: string | string[];
-	entityIds: string | string[];
+	entityId: string;
```
shawnmcknight, line 17: This was never returning multiple `entityIds` and would have caused a problem in the consumer (see comment in `Document` class), so I made it a single always.

src/ts/shared/schemaType/ArrayType.ts
```diff
@@ -0,0 +1,45 @@
...
+
+	/** Create an array of foreign key definitions that will be validated before save */
+	public override transformForeignKeyDefinitionsToDb(value: unknown): ForeignKeyDbDefinition[] {
+		return ensureArray(value)
+			.map((itemValue) => this.valueSchemaType.transformForeignKeyDefinitionsToDb(itemValue))
+			.flat();
+	}
+
+	/** Validate the array */
+	public async validate(value: unknown, document: GenericObject): Promise<string[]> {
```
shawnmcknight, line 36: There was validation being run against the array itself, but no validators existed, which led to a codepath which couldn't be tested.  I dropped that.

src/ts/shared/schemaType/BaseDateType.ts
```diff
@@ -0,0 +1,11 @@
+import BaseScalarType from './BaseScalarType';
+
+/** Base Type for all date types */
+abstract class BaseDateType extends BaseScalarType {
```
shawnmcknight, line 4: Now an abstract class

src/ts/shared/schemaType/BaseScalarArrayType.ts
```diff
@@ -0,0 +1,26 @@
+import { ensureArray } from '#shared/utils';
+import type BaseScalarType from './BaseScalarType';
+import BaseSchemaType from './BaseSchemaType';
+
+/** Abstract Base Type for all scalar arrays */
+abstract class BaseScalarArrayType extends BaseSchemaType {
```
shawnmcknight, line 6: Now an abstract class

src/ts/shared/schemaType/BaseScalarType.ts
```diff
@@ -0,0 +1,165 @@
...
+	| SchemaTypeDefinitionISOCalendarDateTime
+	| SchemaTypeDefinitionISOCalendarDate
+	| SchemaTypeDefinitionISOTime
+	| SchemaTypeDefinitionNumber
+	| SchemaTypeDefinitionString;
+
+const ISVALID_SYMBOL = Symbol('Is Valid');
+
+/** Abstract Scalar Schema Type */
+abstract class BaseScalarType extends BaseSchemaType {
```
shawnmcknight, line 35: Now an abstract class and renamed from `SimpleType`

src/ts/shared/schemaType/BaseSchemaType.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export interface SchemaTypeDefinitionBase {
+	path: `${number}` | `${number}.${number}` | `${number}.${number}.${number}` | number;
+	dictionary?: string;
+	required?: boolean;
+	encrypted?: boolean;
+}
+
+/** Abstract Base Schema Type */
+abstract class BaseSchemaType {
```
shawnmcknight, line 12: Now an abstract class and renamed from `BaseType`

src/ts/shared/schemaType/EmbeddedType.ts
```diff
@@ -0,0 +1,63 @@
...
+		subrecord.forEach((value, arrayPos) => {
+			if (typeof value !== 'undefined') {
+				setIn(record, [arrayPos], value);
+			}
+		});
+		return record;
+	}
+
+	/** Validate the embedded document */
+	public async validate(document: Document): Promise<string[]> {
```
shawnmcknight, line 51: Similarly here this was returning `Object[]` which was different than the base class, so it was adjusted.

src/ts/shared/schemaType/ISOCalendarDateTimeType.ts
```diff
@@ -0,0 +1,111 @@
...
+	public constructor(
+		definition: SchemaTypeDefinitionISOCalendarDateTime,
+		options: ScalarTypeConstructorOptions = {},
+	) {
+		super(definition, options);
+
+		const { dbFormat = 'ms' } = definition;
+		this.isDbInMs = dbFormat === 'ms';
+
+		this.isoCalendarDateType = new ISOCalendarDateType(
```
shawnmcknight, line 35: I made these instance members rather than instantiating on the fly during validation because they required a definition to avoid type errors.  I think this is a better pattern anyway.

kthompson23 (approved):
This looks really solid. The tests were easy to follow too.

PR #18, opened 2022-03-08
Convert Connection, Query, and compileModel to TS

shawnmcknight:
This PR converts the `Connection` and `Query` classes as well as the `compileModel` function from JavaScript to TypeScript.  Other than the entrypoint index, this completed the conversion of all the packaged JavaScript (some unpackaged script files still remain and will be converted later).

Since there is basically no JavaScript remaining and all existing tests have been rewritten, much of the existing dev tool configuration was able to be streamlined.  `babel-plugin-rewire-ts` was eliminated from the babel configuration and the configuration was updated to handle the existing environment and file structures.  Some eslint overrides for javascript were eliminated.  Similar changes were made to the jest config.  Coverage requirements were increased back to 100%.  Unused dependencies, including `sinon` were dropped.

This work allowed the input and output from the Db server subroutines to be typed which is a huge improvement in interacting with that layer.

During the conversion, some bugs were uncovered and corrected and some code was refactored to improve its structure.  I will attempt to highlight those changes in comments inline with the code.


src/js/index.js
```diff
@@ -1,25 +1,6 @@
...
-	 * @member {Schema} Schema
-	 * @memberof mvom
-	 */
-	Schema,
-	createConnection,
-};
-
-export default mvom;
-export { Document, Errors, Schema, createConnection };
+export { Document, Errors, Schema, Connection };
```
shawnmcknight, line 6: I changed this export pretty significantly which will potentially be a breaking change when the release is created.  I see no reason to have an default export of an object as well as named exports.
Additionally, `createConnection` was eliminated as a standalone factory function and instead is now a static factory method on `Connection`

src/ts/Document.ts
```diff
@@ -52,8 +50,6 @@ class Document {
 	/** Indicates whether this document is a subdocument of a composing parent */
 	private readonly isSubdocument: boolean;
 
-	public constructor(schema: Schema, options: DocumentConstructorOptions);
```
shawnmcknight, line 55: I had to undo these changes.  It turned out with how `Model` calls `super` here that it couldn't determine if there were appropriate parameters.

src/ts/Document.ts
```diff
@@ -36,8 +36,6 @@ export interface BuildForeignKeyDefinitionsResult {
 class Document {
 	[key: string]: unknown;
 
-	public _id?: string;
```
shawnmcknight, line 39: This is a little wonky and might need further refactor later.  A `Document` doesn't really have an `_id`, but a `Model` instance does.  However, there are multiple places in the `Document` class looking at the `_id`.  For now it'll just become part of the index signature and type checking will need to be more robust around it.

src/ts/shared/errors/DbServerError.ts
```diff
@@ -1,27 +1,31 @@
...
+type DbServerErrorConstructorOptions =
+	| DbServerErrorConstructorOptionsMessage
+	| DbServerErrorConstructorOptionsCode;
+
 /**
  * Error thrown when an error is encountered on the database server
  */
 class DbServerError extends BaseError {
-	public constructor({ message: overrideMessage, errorCode }: ConstructorOptions) {
+	public constructor(options: DbServerErrorConstructorOptions) {
```
shawnmcknight, line 20: There were two different kinds of calls into this error.  One with just a message, and another with a code provided.  I had to change the definition to support both.

src/ts/Query.ts
```diff
@@ -0,0 +1,305 @@
...
+	/** Limit the result set to this number of items */
+	private limit?: number;
+
+	/** Skip this number of items in the result set */
+	private skip?: number;
+
+	/** Specify the projection attribute in result set */
+	private projection: string[];
+
+	public constructor(
```
shawnmcknight, line 86: Originally this class allowed for modifications of the parameters on the fly (e.g. limit) via dedicated methods.  We've never used them and the `Query` class is never directly exposed to consumers.  This was really a holdover from trying to model this like mongoose.  I simplified things significantly by removing that logic.

src/ts/Query.ts
```diff
@@ -0,0 +1,305 @@
...
+			throw new InvalidParameterError({
+				message: 'Nonexistent schema property or property does not have a dictionary specified',
+				parameterName: 'property',
+			});
+		}
+		return dictionaryId;
+	}
+
+	/** Transform query constant to internal u2 format (if applicable) */
+	private transformToQuery(property: string, constant: unknown): unknown {
```
shawnmcknight, line 293: This method did not exist previously.  A method which returned the transformer method was in its place.  Rather than having the caller get the method and then run it, it seemed better just to run the transform.

src/ts/Query.ts
```diff
@@ -0,0 +1,305 @@
...
+			queryCommand,
+			...(this.skip != null && { skip: this.skip }),
+			...(this.limit != null && { limit: this.limit }),
+			projection,
+		};
+
+		this.Model.connection.logger.debug(`executing query "${queryCommand}"`);
+		const data = await this.Model.connection.executeDbFeature('find', options);
+
+		return {
```
shawnmcknight, line 125: Previously this would return `Model` instances by calling a method on the `Model` class.  I figured it would be better to just let the `Model` class do that itself.

src/ts/Connection.ts
```diff
@@ -0,0 +1,481 @@
...
+	) {
+		this.logger = logger;
+		this.cacheMaxAge = cacheMaxAge;
+		this.endpoint = `${mvisUri}/${account}/subroutine/${Connection.getServerProgramName('entry')}`;
+		this.timeout = timeout;
+
+		logger.debug(`creating new connection instance`);
+	}
+
+	public static createConnection(
```
shawnmcknight, line 138: This is now a static factory method rather than a standalone method.  The constructor is private so the only way to instantiate this should be via the factory.

src/ts/Connection.ts
```diff
@@ -0,0 +1,481 @@
...
+	 * Create directory when deploying features
+	 * @defaultValue false
+	 */
+	createDir?: boolean;
+}
+
+type ServerDependency = keyof typeof serverDependencies;
+
+interface ServerFeatureSet {
+	validFeatures: Map<ServerDependency, string>;
```
shawnmcknight, line 81: This was changed from an object.

src/ts/Connection.ts
```diff
@@ -0,0 +1,481 @@
...
+	invalidFeatures: Set<ServerDependency>;
```
shawnmcknight, line 82: This was changed from an array.

src/ts/Query.ts
```diff
@@ -0,0 +1,305 @@
...
+	/** Greater than or equal to */
+	$gte?: TValue;
+	/** Less than */
+	$lt?: TValue;
+	/** Less than or equal to */
+	$lte?: TValue;
+	/** Not equal */
+	$ne?: TValue;
+	/** String containing */
+	$contains?: TValue;
```
kthompson23, line 32: Does it make sense to type $contains, $startsWith, and $endWith as string?

shawnmcknight: Of the scalar types, I can think of `string` and `number` as usable with these operators, although I would think that there would be a very limited use case for `number`.  The date types and the boolean type wouldn't make any sense.

I'm not sure its correct to change this here.  It's a little awkward in general that you _could_ use those operators on a property which isn't defined as a string at all.  It might be more awkward to force someone to provide a string for something that isn't a string were they to try to use the operator.

I did have a thought here though after looking at the MongoDB node types.  Perhaps these should be `TValue extends string ? string : never;`  That would probably be the best path because it would prevent a user from using it wrong.  I'll make that change.

image

kthompson23 (approved):
Looks great. I like all the new type safety.

PR #20, opened 2022-03-09
Final conversion of repo from JS to TS

shawnmcknight:
This PR completes the conversion of the repo from JavaScript to TypeScript.  Included in this PR:

* Remove files from nested `src/ts` folder to the root `src` folder
* Eliminate the `#shared` alias from tsconfig, babel config, and eslint config.  This is because declaration file generation does not work with the aliased paths, and the benefit of that outweighs the benefit of the aliasing.
* Elimination of the `src/shared` folder in favor of `src/group` folders.
* Add a declaration file generation step to the build process
* Change the pointer to the main index and the declaration file in `package.json`
* Convert build scripts from js to ts and refactor to improve their structures
* Use js private field and method syntax in favor of ts private directive to ensure that delcaration file could be built for `Model` and its parent `Document`.  TS will not build the declaration file for members declared as private which are part of a class returned from a function.
* Allow skip/limit in `Query` constructor to accept `null`
* Update the main index to export useful types

Although there are a lot of files in this PR, most of it is renames along with changes to use relative pathing.


src/compileModel.ts
```diff
@@ -65,7 +65,7 @@ const compileModel = <TSchema extends GenericObject = GenericObject>(
 		public readonly __v: string | null;
 
 		/** Private id tracking property */
-		private __id: string | null;
+		#__id: string | null;
```
kthompson23, line 68: Is this the only place where we were prefixing private members with an underscore? With the private keyword it doesn't seem necessary anymore.

kthompson23 (approved):
Looks awesome.

src/compileModel.ts
shawnmcknight: I agree about not needing the underscore prefix any longer, but we probably have to be a little careful with `Model`/`Document` instances since the properties of the schema get merged into it.  I suppose no identifier is truly safe from being used by a consumer (short of throwing an error in the `Schema` constructor if they get used), but hopefully something prefixed with two underscores will avoid a potential clash.

I believe I removed the underscore from everywhere otherwise, with the exception of `_id` , `__v`, and `_raw` which are named as such by convention.

kthompson23: The nice thing about the use of the private identifier is that `this.#_id` and `this._id` can coexist so it can't be accidentally clobbered.

shawnmcknight: Hmm, I think I knew that, but didn't really think about it.  I will convert it over in my other open PR.

PR #24, opened 2022-03-11
Drop support for node 12

shawnmcknight:
Node 12 is EOL is a little more than a month.  Since the next release of this library will be a breaking change anyway, it makes sense to drop support for Node 12 at this time.   This PR adjusts the following to eliminate support for Node 12:
* `package.json` engines is now minimum `14.19`
* Babel compilation target is node `14.19.0`
* CI workflow adjusted to no longer run against node@12


kthompson23 (approved):
Looks good.

PR #26, opened 2022-03-11
Ensure that the prettier property is dropped from dist package.json

shawnmcknight:
During the last publish I noticed that the `dist` folder's `package.json` still contained the `prettier` property which is unnecessary for the published version.  This PR updates the `copy-files` script to ensure that it does not get copied into the published version.


kthompson23 (approved):
nice.

PR #29, opened 2022-03-14
Fix encryption type and eliminate unions with unknown

shawnmcknight:
While working on integrating the 2.0.0-alpha.0 version, I found that the types for encryption and decryption were not working as expected.  This is because whenever a union type includes `unknown` is gets narrowed to `unknown` (i.e. the other types in the union are removed).  That made implementing an array variant of the encryption/decryption functionality impossible to get typed properly.

This PR makes a few changes:
* The encryption and decryption types have been adjusted to only allow strings for both input and output.
  * By the time the encryption functionality runs, all scalar types have already been converted to a string representation of that scalar.  Therefore, supporting any other input types is not necessary.
  * Rather than expecting consumers to transform `null`, and array types, this logic has been internalized into the encryption and decryption functionality.  This also addresses a likely defect in the handling of nested arrays which were unlikely to have been accounted for by consumers.
  * For decryption, it should be anticipated that any encrypted value would already be a string, and if it isn't, then it shouldn't be decrypted.  Since casting to the appropriate scalar type occurs after decryption, it should be fine to output a string and then let the scalar definition cast it to the appropriate data type.
  * This is a breaking change because of the type definition changes.
* The types for the shape of an `MvRecord` have been improved.  The new types ensure that an attribute cannot be an infinite structure of nested arrays, and also introduces two new types `MvDataType` and `MvAttribute` to make working with those structures easier to type.
* All unions with `unknown` have been eliminated.  This required some ancillary implementation detail changes to make happen.
* A bug that occurs on the date of daylight savings time was uncovered but only some of the tests dealing with time conversions illustrated it.  This bug occurs because the start of the day as compared to the current time is shifted by one hour on the day of daylight savings beginning or ending if the time is later than the shift.  Although this bug wasn't fixed in this PR, the tests were adjusted so that they would consistently fail.  Some of the tests were passing because the time was earlier than the transition.


kthompson23 (approved):
Looks good.

PR #31, opened 2022-03-14
Fix _id enumerability on Model

shawnmcknight:
The `_id` property on the `Model` instance has a getter and setter to prevent overwrite of the id.  This is using es6 style getters and setters which are defined on the instance's prototype rather than the instance itself which prevents it from being enumerable.  This property needs to be enumerable so the es6 getters and setters were replaced with `Object.defineProperty` getters and setters.


kthompson23 (approved):
Interesting.

PR #37, opened 2022-03-16
Return null for missing records, fix path for UniBasic source, and loosen Document constructor options

shawnmcknight:
This PR fixes the following issues:

- The database layer previously returned empty string for missing records in the `findById` and `findByIds` features.  The response from the DB for these features was typed as `null` rather than empty string.  As a result, logic in the `Model` which was previously returning `null` was instead returning that empty string causing downstream errors.  To resolve this, those DB features have been adjusted to align with the type and will now return `null` from the database.  This is a breaking change to those features to the version number has been bumped to the next semver major as part of this PR.
- The path for unibasic source was incorrect as it was still looking to the parent directory for the `unibasic` folder.  When the paths were flattened during the TS conversion this was not taken into account.  The path has been corrected so unibasic source can be deployed properly.
- The `Document` and `Model` constructors were requiring that only `data` or `record` options be specified.  It is actually valid syntax to specify both `data` and `record` properties.  When this syntax is used, the record will be applied first and the data will be applied afterwards.  The constructor options have been loosened to allow for both to be specified at the same time.


src/.mvomrc.json
```diff
@@ -4,8 +4,8 @@
     "deploy": "^1.0.0",
     "entry": "^0.2.0",
     "find": "^1.0.0",
-    "findById": "^1.0.0",
-    "findByIds": "^1.0.0",
+    "findById": "^2.0.0",
+    "findByIds": "^2.0.0",
     "readFileContentsById": "^1.0.0",
```
kthompson23, line 9: `readFileContentsById` can also return an empty string. Should it return null?

kthompson23: readFileContentsById is typed to always return a string so it's probably okay.

PR #42, opened 2022-03-18
Replace moment with date-fns

shawnmcknight:
This PR eliminates the `moment` dependency in favor of `date-fns`.  `moment` is a legacy project and is in maintenance mode and they recommend using a different library while `date-fns` is an active project.

Most places where `moment` was used previously have been replaced with comparable functions from `date-fns`.  An exception to this is in the `IsoTimeType` schema type formatters.  A separate bug was discovered where the math operation being used to convert to/from internal time to ISO time was working incorrectly on the date that a daylight savings time shift occurred.  This was because the previous calculation was relying on the difference between the start of the day and the value being converted.  On days where daylight savings time shifts, this caused the time to be off by +/- 1 hour.  To correct this problem, the calculation for converting to/from internal time was internalized.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,9 +10,9 @@
       "license": "MIT",
       "dependencies": {
         "axios": "^0.26.1",
+        "date-fns": "^2.28.0",
         "fs-extra": "^10.0.1",
         "lodash": "^4.17.21",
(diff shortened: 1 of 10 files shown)


kthompson23 (approved):
This looks good.

PR #43, opened 2022-03-18
Make timeout errors explicit

shawnmcknight:
This PR adds explicit handling of "timeout" errors on Axios requests and throws a new `TimeoutError` if a requests times out.  This will allow MVOM to distinguish between other errors that occur on the request versus errors due to timeouts on the request.

There has also been some refactoring on the connection class to make future changes in this area easier to implement:
* Rather than `axios.post` being called with the request options specified each time, a new `AxiosInstance` is created when the `Connection` is instantiated.  This means configuration of things like the timeout value, the url, and others only needs to happen a single time.
* Error handling overall has been improved.  When an error is thrown during a request, it will be evaluated to see if it is an `AxiosError` and if so, it will be handled specifically as that type of error.  This allows things like the aforementioned checking for timeouts as well as potentially some future enhancements to error detection.  If an error is not an axios error, a new `UnknownError` instance will be thrown.  This should hopefully be an edge case.


src/Connection.ts
```diff
@@ -137,8 +136,14 @@ class Connection {
...
 		this.cacheMaxAge = cacheMaxAge;
-		this.endpoint = `${mvisUri}/${account}/subroutine/${Connection.getServerProgramName('entry')}`;
-		this.timeout = timeout;
+
+		const baseURL = `${mvisUri}/${account}/subroutine/${Connection.getServerProgramName('entry')}`;
+
+		this.axiosInstance = axios.create({
+			baseURL,
+			timeout,
+			transitional: { clarifyTimeoutError: true },
```
kthompson23, line 145: Any idea why this is "transitional" and might go away in future versions? Do they expect you to parse the timeout message?

kthompson23 (commented):
Looks good, just a question about the "transitional" options.

src/Connection.ts
shawnmcknight: I didn't find anything when I originally looked, but now I see this in the Github README:
```
// transitional options for backward compatibility that may be removed in the newer versions
```
So it seems likely that at some stage there will be a breaking change.  That might mean we can't ascertain this anymore -- it might mean there's an API change, it might mean they always send the timeout error in this situation without the flag.  I can't find anything else about the intent on it.  I guess when updating the version some care will need to be exercised.

PR #44, opened 2022-03-18
Improve logging behavior

shawnmcknight:
This PR updates the logging behavior by changing the following:
* All log messages will now be prefixed by the account name in brackets (i.e. `[ACCOUNT_NAME] log message`)
* Log levels of specific messages were given better semantics (i.e. errors now log at error level, connection information at info level, etc.)
* Additional logging was added for many error conditions


kthompson23 (approved):
This looks great.

PR #50, opened 2022-03-22
Pass dbformat to ISOTime constructor

shawnmcknight:
This PR fixes a bug in the `ISOCalendarDateTime` schema type when there is no `dbFormat` property specified.  The `ISOCalendarDateTime` type defaults to a format of `ms` (milliseconds) when no `dbFormat` is provided.  Currently, when it instantiates an instance of the `ISOTime` schema type it is only passing the provided definition to that constructor which means that the defaulted format is not provided.  The `ISOTime` schema type defaults to a format of `s` (seconds).  This causes an error in the process of transforming the data.

The change is to ensure that the `dbFormat`, including the defaulted `ms` variant, are passed through to the `ISOTime` constructor.


kthompson23 (approved):
Looks good to me.

PR #53, opened 2022-03-23
Allow type information for dictionaries

shawnmcknight:
This PR introduces a new feature to allow for specification of type information for additional dictionaries provided as options to the `Schema` constructor.  Currently, dictionaries can be added through these options to allow queries to be executed against dictionaries not defined in the schema, such as virtual dictionaries.  However, the constants provided for these dictionaries cannot be transformed into multivalue format because there is no type information.  For instance, if a dictionary resolves to a multivalue boolean (`1` or `0`) then the consumer must provide a `1` or a `0` as the query conditional's constant value rather than the more familiar `true` or `false`.  With the addition of type information, the query constants will be transformed to multivalue internal format in the same manner as dictionaries defined in the schema definition.

In order to allow for backwards compatibility, the existing format of providing dictionaries as key/value pairs representing a property and a dictionary was retained.  When using this syntax, the constants will be treated as strings and the only casting done will be to convert any input value to a string.  This should match the existing behavior.  It is now additional possible to provide an object in place of the dictionary name and supply a `dictionary` property (which will contain the dictionary ID) and a `type` property (which will match the type names of the schema definition scalars).  Additionally, the `ISOCalendarDateTime` and `ISOTime` types will allow for a `dbFormat` to be specified to indicate if the time is in seconds or milliseconds, similar to the schema type option.

The implementation of this change involved the introduction of a new `DataTransformer` interface.  The data transformation logic resident in each schema type was then moved to a class implementing this interface which only facilitates the data transformation logic and does not offer the other logic resident in the schema type.  The scalar schema types were adjusted to also implement this same interface, and the existing transformation methods now delegate to the corresponding `DataTransformer` class implementation.  When adding dictionaries with type information, an appropriate `DataTransformer` will be instantiated and used for data transformations for those.  Dictionaries defined by schema types will continue to use the transformation methods in the schema types, but as previously stated those merely delegate to another `DataTransformer` implementation.  This allows for both schema types and data transformers to perform the transformations with a common interface.


src/dataTransformers/NumberDataTransformer.ts
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Number Data Transformer
+ * @throws {@link InvalidParameterError} An invalid parameter was passed to the function
+ */
+class NumberDataTransformer implements DataTransformer {
+	/** Number of implied decimals in database storage */
+	private readonly dbDecimals: number;
+
+	public constructor(dbDecimals = 0) {
+		if (!Number.isInteger(dbDecimals)) {
```
kthompson23, line 13: This should also validate that dbDecimals is a positive interger

src/Schema.ts
```diff
@@ -137,7 +173,64 @@ class Schema {
...
+			switch (type) {
+				case 'string':
+					return acc.set(queryProperty, {
+						dictionary,
+						dataTransformer: new StringDataTransformer(),
+					});
+				case 'number':
+					return acc.set(queryProperty, {
+						dictionary,
+						dataTransformer: new NumberDataTransformer(),
```
kthompson23, line 206: Can't you specify an oconv value in a virtual dictionary?

kthompson23 (commented):
Looks good. A made a note about validating that dbDecimals is a positive number and a question about the use of oconv for numeric virtual dictionaries.

src/Schema.ts
shawnmcknight: Maybe I'm not quite following the question, but the number query transformer doesn't do anything other than return the string representation of whatever value was provided, just like the string query transformer.  The expectation is that the query engine would handle the decimals.

src/dataTransformers/NumberDataTransformer.ts
shawnmcknight: Good call.  This wasn't in the original schema type but it makes sense to add.

src/Schema.ts
shawnmcknight: @kthompson23 Not quite sure if there's an issue that you're concerned about or not.  Let me know if there's something.

kthompson23: OK that makes sense.

kthompson23 (approved):

PR #60, opened 2022-03-31
Change database I/O for records to strings instead of arrays

shawnmcknight:
This PR changes the mechanism for transferring records to and from the database to use multivalue delimited strings instead of arrays.  There are a few underlying reasons for this change:

1. The UniBasic code to convert the dynamic array to a UDO Array is very complicated and challenging to debug and modify.  The equivalent TypeScript code is significantly easier to understand, extend, and debug.
2. This should end up as a performance improvement.  Currently a lot of work is done on the DB backend to turn the dynamic array into a UDO array, only to then serialize it into JSON which then also incurs its own performance penalty on large data sets.  The DB server is not scalable in the same way that a TypeScript consumer might be without vertically scaling the DB server and a TypeScript consumer can likely be horizontally scaled.  I believe avoiding the work of converting to an array, then serializing the array for transit and serializing it back into an array will improve things significantly.
3. By handling multivalue strings directly, `Document` instances can be constructed directly from a multivalue string.  This will allow for multivalue strings, regardless of how they have been acquired, to be converted into a `Document`.  Therefore it becomes possible to get data outside of the DB connection created through this library and still apply a schema to it and construct an object to work with.  Essentially this increases the flexibility of the library for future use cases that might not require database connections.

The main changes in this PR are to adjust the `formatDocument` partial and the `save` database feature to emit and receive multivalue delimited strings.  Additionally, the `getServerInfo` database feature has been adjusted to emit the delimiter marks in use on the database server so the strings can be appropriately split regardless of which delimiters are being used by the database server.  Defaults are applied prior to connection, but will be overwritten once the server connection has been established.

The `Model` constructor has been adjusted to accept a delimited string instead of an array and the `Document` class has been augmented with several static methods to create `Document` instances via specific factory methods instead of via the constructor.  The constructor has been marked as `protected` to encourage consumers to use the factory methods instead of the constructor.  One of those factory methods will allow for instantiating a `Document` via a multivalue string for the future use case indicated in point 3 above.  Several internal changes were made to accommodate for these changes to the constructors.

Finally, because the instantiation of `Model` instances has been changed to use a string instead of an array, the use case where an array could be provided to the `Model` constructor to act as a template has been removed and is a breaking change.  To ease consumers ability to navigate this change, a new non-enumerable property `_originalRecordString` has been added to the `Model` instance which will retain the string that the model was instantiated from.  If there is a desire to provide a template for a new `Model` instance, this string can then be passed in to the `Model` constructor and it will act as the baseline for the data on newly created instances.


src/Document.ts
```diff
@@ -36,25 +36,81 @@ class Document {
...
+		recordString: string,
+		dbServerDelimiters: DbServerDelimiters,
+	): Document {
+		const record = Document.convertMvStringToArray(recordString, dbServerDelimiters);
+
+		return new Document(schema, { record });
+	}
+
+	/** Convert a multivalue string to an array */
+	public static convertMvStringToArray(
```
shawnmcknight, line 79: Ideally this would have been `protected` as there is no use case for consumers outside this class.  However, there is an open TS issue where `private` and `protected` cannot be used in class constructors that are assigned to variables/returned from functions.  Therefore, this has been made `public` although there is no known external use case.

shawnmcknight:
I've converted this to a draft as there's some tweaks to how the delimiters are being ascertained I'd like to make.

src/Document.ts
```diff
@@ -36,25 +36,81 @@ class Document {
...
+		const { am, vm, svm } = dbServerDelimiters;
+		const record: MvRecord =
+			recordString === ''
+				? []
+				: recordString.split(am).map((attribute) => {
+						if (attribute === '') {
+							return null;
+						}
+
+						const attributeArray = attribute.split(vm);
```
kthompson23, line 92: If an attribute just contains subvalues we'll never hit the below condition to further split on svm.

kthompson23 (commented):
I'm concerned that the method to retrieve the delimiters will not work when running in language group 2 or 3 because those characters are unprintable.

I also think there is a bug where attributes that just contain subvalues will not be converted correctly.

src/Document.ts
shawnmcknight: I've adjusted this to test the string for inclusion of svm marks and continue if they are found.  Unit tests have been added to  test this scenario as well.

src/unibasicTemplates/getServerInfo.njk
```diff
@@ -12,6 +14,36 @@ if udoSetProperty(output, 'time', system(12)) then
   go returnFromSub
 end
 
+if udoCreate(UDO_OBJECT, delimiters) then
```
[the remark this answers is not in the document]
shawnmcknight: After research, it appears that JSON can accept any character in the range greater than or equal to hex 0020.  For characters below 0020, which includes the language group 3 delimiters, UDO will emit the Unicode point in format `\unnnn` (e.g. `\u0014`).  Everything seems to work fine with this.

kthompson23 (approved):
This looks great.

PR #76, opened 2022-04-12
Allow user defined properties in Model methods

shawnmcknight:
This PR allows for custom user defined properties to be passed to the Model methods as a property of an object named `userDefined`.  This object can contain up to 5 properties named `option1` , ..., `option5`.  These properties will be passed to the database subroutines and added to the existing `S$MVOM` named common block as `S$MVOM.USER1`, ..., `S$MVOM.USER5`.

These common variables then can be used in upstream functions such as database triggers.

One additional change was made as well with the `.mvomrc.json` file being renamed as `manifest.json` which is a more appropriate term to use for this file.


src/compileModel.ts
```diff
@@ -164,12 +182,16 @@ const compileModel = <TSchema extends GenericObject = GenericObject>(
...
-			});
+			const { projection = [], userDefined } = options;
+			const data = await Model.connection.executeDbFeature(
+				'findById',
+				{
+					filename: Model.file,
+					id,
+					projection: Model.schema?.transformPathsToDbPositions(projection) ?? [],
+				},
+				userDefined && { userDefined },
```
kthompson23, line 193: Is there any use for these values when calling `find`? No non-mvom code can react to it. I suppose we could implement some kind of server side logging in the future where we could log these user defined values alongside whatever else we decide we want to log?

I wouldn't be opposed to only passing these to the save and delete routines.

src/unibasicTemplates/setup.njk
```diff
@@ -3,7 +3,37 @@ subroutine mvom_setup(setupOptions)
 * Perform any needed environment setup
 
 com /S$MVOM/ S$MVOM.PROCESS
+com /S$MVOM/ S$MVOM.USER1, S$MVOM.USER2, S$MVOM.USER3, S$MVOM.USER4, S$MVOM.USER5
```
kthompson23, line 6: You can declare on multiple lines like this?

kthompson23 (commented):
Looks good but I'm wondering about the need to pass user defined values to anything other than delete and save. Also, just making sure that the multiline named common block was intentional.

src/unibasicTemplates/setup.njk
shawnmcknight: Yes.  I wasn't aware of it, but it works fine.  The only rule apparently is that the variable names are by ordinal position and not by name.  Each additional statement just adds more positions to the list.

src/compileModel.ts
shawnmcknight: I can't think of a use case for anything other than save and delete.  I don't know of any way to access them upstream except in triggers, and triggers are only fired on update and delete operations.  I guess if we ever added additionally functionality for these variables other than for adding to common it might come in handy.  I can remove it if you think its superfluous.

kthompson23: I suppose it's harmless.

kthompson23 (approved):
Looks good to me.

PR #79, opened 2022-04-15
Validate server limits for queries

shawnmcknight:
This PR adjusts the query process to validate queries prior to execution to ensure they do not exceed the limits established by the database server.  The `getServerInfo` database feature will execute the `LIMIT` verb, parse the output, and return the maximum sentence length, with conditions, and sort conditions which can be in a query.  These values will be used to compare against the query that is to be executed, and if there is a violation of the server limits then a new `QueryLimitError` will be thrown and the query will not be executed.


src/unibasicTemplates/getServerInfo.njk
```diff
@@ -4,12 +4,12 @@ subroutine mvom_getServerInfo(options, output)
 
 {% include "./partials/errorConstants.njk" %}
 
-if udoSetProperty(output, 'date', date()) then
+if udoSetProperty(output, 'date', date(), UDO_NUMBER) then
```
shawnmcknight, line 7: These values were typed as number for the subroutine output, but the db server was actually returning strings.  This didn't seem to cause any issues, but this change makes the output consistent with expectations.

src/unibasicTemplates/getServerInfo.njk
```diff
@@ -44,6 +44,44 @@ if udoSetProperty(output, 'delimiters', delimiters) then
   go returnFromSub
 end
 
+if udoCreate(UDO_OBJECT, limits) then
+  call error_handler(ERROR_UDO, output)
+  go returnFromSub
+end
+
+hush off
```
shawnmcknight, line 52: "hush", which is enabled in the entry point, had to be disabled in order to get the `CAPTURING` clause to work here.

src/unibasicTemplates/find.njk
```diff
@@ -23,14 +23,6 @@ if udoGetProperty(options, 'projection', projection, type) then
   go returnFromSub
 end
 
-* Temporary length check on the query since this is the query limit we are most likely to exceed.
-* STORIS Rocket case 00656416 - exceeding a query limit does not set an error condition in @system.return.code
-* TODO CXM-2318 - Add handling to mvom to prevent queries that will exceed db server limits
```
kthompson23, line 28: There's another reference to CXM-2318 below (can't stick a comment there for some reason).

With this change we should be able to re-enable using the `returning` clause so we can return better error codes.

kthompson23 (commented):
The changes look good. There's one additional reference to CXM-2318 in find.njk though

src/unibasicTemplates/find.njk
shawnmcknight: It looks like it was using the `returning` statement previously but wasn't doing anything with the error code result.  I reverted it back to the way it was previously, but there might need to be additional work to do something with the error codes.

kthompson23 (approved):
Looks good.

PR #85, opened 2022-04-22
Fix improper saving of subdocument arrays with multi-part paths

shawnmcknight:
This PR fixes a bug when using schemas defined with document arrays that have paths defined at the value position. Consider the following schema definition:

```ts
const definition: SchemaDefinition = {
  subdocumentArray: [
    {
      prop1: { type: 'string', path: '1.1' },
      prop2: { type: 'number', path: '1.2', dbDecimals: 2 },
    },
  ],
};
```

Using this schema, if you had a document with a property of `subdocumentArray` that was:
```
[ { prop1: 'foo', prop2: 1.23 }, { prop1: 'bar', prop2: 4.56 }]
```
this should result in a multivalue record of:
```
foo{svm}bar{vm}123{svm}456
```
However, the current code is saving this as:
```
foo{svm}123{vm}bar{svm}456
```

This PR changes the handling of how subdocuments are set into data to ensure that multi-part paths are properly handled when saving subdocuments.

### DIFF
--- a/src/__tests__/Document.test.ts
+++ b/src/__tests__/Document.test.ts
@@ -413,6 +413,30 @@ describe('createDocumentFromRecordString', () => {
 
 		expect(document.prop1).toEqual([['foo', 'bar', 'baz']]);
 	});
+
+	test('should create a new document from the provided record string using document arrays at subvalue positions', () => {
+		const definition: SchemaDefinition = {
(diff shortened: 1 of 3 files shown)


kthompson23 (approved):
nice, looks good.

PR #86, opened 2022-04-22
Allow projection of only ids

shawnmcknight:
Currently, the behavior of the `projection` option (`find`, `findAndCount`, `findById`, `findByIds`) is such that if the consumer were to provide an empty array of projection values then the behavior would be identical to providing no projection option at all.  Only if there are one or more values in the `projection` array will projection change the output result

This behavior is unexpected as an empty projection array should indicate that nothing should be projected at all.  This change adjusts the behavior so that no _data_ will be projected if an empty array is provided for `projection`.  The `_id` and `__v` values will continue to be emitted from the db server, however, and will be valued in the resulting document.

### DIFF
--- a/src/compileModel.ts
+++ b/src/compileModel.ts
@@ -124,10 +124,10 @@ const compileModel = <TSchema extends GenericObject = GenericObject>(
 		): Promise<Model | null> {
 			const { userDefined } = options;
 
-			const data = await Model.connection.executeDbFeature(
+			const data = await this.connection.executeDbFeature(
 				'deleteById',
 				{
-					filename: Model.file,
+					filename: this.file,
 					id,
 				},
 				userDefined && { userDefined },
@@ -139,7 +139,7 @@ const compileModel = <TSchema extends GenericObject = GenericObject>(
 
 			const { _id, __v, record } = data.result;
 
-			return Model.#createModelFromRecordString(record, _id, __v);
+			return this.#createModelFromRecordString(record, _id, __v);
 		}
 
 		/** Find documents via query */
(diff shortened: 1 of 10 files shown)


src/unibasicTemplates/findByIds.njk
```diff
@@ -39,6 +33,12 @@ if recordIdsCount eq 0 then
   go returnFromSub
 end
 
+* get the projections from the options
```
shawnmcknight, line 36: This was an existing bug that was exposed by changing projection to be an array or null.  There is a type check above this block that is intended to validate that the list of ids is an array.  However, when projection got added, it used the same `type` variable that the `ids` property get did, which means the type validation was running against the projection array rather than the ids array.  I simply relocated this `udoGetProperty` call to ensure that the proper `type` variable is being inspected.

src/compileModel.ts
```diff
@@ -124,10 +124,10 @@ const compileModel = <TSchema extends GenericObject = GenericObject>(
 		): Promise<Model | null> {
 			const { userDefined } = options;
 
-			const data = await Model.connection.executeDbFeature(
+			const data = await this.connection.executeDbFeature(
```
kthompson23, line 127: Are these interchangeable in JS?

kthompson23 (approved):
Other than the question about switching to `this` in the Model class this looks good.

src/compileModel.ts
shawnmcknight: In a static member, `this` references the other static members, so using the class name isn't necessary.  I'm not sure if this wasn't always the case or if I misunderstood it until recently: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static#calling_static_members_from_another_static_method

PR #105, opened 2022-05-04
Add babel runtime to dependencies

shawnmcknight:
The `@babel/plugin-transform-runtime` transformation was added to the transpiled output several releases ago, but the required `@babel/runtime` production dependency was never added alongside it.  If the runtime dependency isn't there, the library will fail to run properly as babel has transformed the output with the expectation that it will be available.

This PR adds `@babel/runtime` to the list of production dependencies.


kthompson23 (approved):
Was this causing actual problems?

PR #106, opened 2022-05-04
Throw with original value instead of partially transformed value

shawnmcknight:
There are scenarios where the transformation of data from multivalue cannot be performed which results in a logging of a `TransformDataError`.  Several of these scenarios where this error is thrown are including the failed transformation value in the error.  This makes diagnosis of problems difficult.  For example, if a number can't be transformed and results in a `NaN` value, the `NaN` will be logged instead of the original value.

This PR changes those logs to record the value prior to any modification so tracing the origin can be easier.


kthompson23 (approved):
Looks good.

PR #128, opened 2022-05-20
Handle trailing slash in mvis uri more gracefully

shawnmcknight:
Currently, if the client passes an `mvisUri` with a trailing slash, the generated `baseURL` for axios will end up with two slashes in it due to the use of templated strings to construct the `baseURL`.  This PR adjusts the logic so that the `baseURL` will be created using the `URL` constructor which will properly build the url with and without a trailing slash.


kthompson23 (approved):
Looks good.

PR #152, opened 2022-06-06
Remove @types/jest and modern option for fakeTimers

shawnmcknight:
#145 attempted to update `@types/jest` but encountered CI failures on typechecking.  Having seen that, I recalled that `jest` ships its own types now and use of `@types/jest` is no longer necessary.  This PR eliminates the `@types/jest` dependency and now jest's internal types are being used.

Additionally, as of [jest 27](https://jestjs.io/blog/2021/05/25/jest-27#flipping-defaults), modern timers are now the default.  I've updated the calls to `useFakeTimers` to no longer pass the `modern` flag.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,6 @@
         "@storis/prettier-config": "1.0.0",
         "@storis/tsconfig": "1.0.0",
         "@types/fs-extra": "9.0.13",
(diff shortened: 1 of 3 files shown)


kthompson23 (approved):
Looks good.

PR #157, opened 2022-06-08
Fix findById use of UDO_NULL

shawnmcknight:
The `findById` BASIC program is missing the include of the UDO constants.  As a result, the use of `UDO_NULL` in the program is uninitialized.  This PR updates the program to include the constants to prevent that problem.


kthompson23 (approved):
Looks good.

PR #196, opened 2022-07-06
Check catalog existence as part of server validation

shawnmcknight:
#186 introduced new functionality to validate and deploy the REST definitions and UniBasic subroutines using MVIS admin.  That functionality relies solely on the existence of a REST definition in MVIS to indicate whether or not the connection is valid and can be opened.  There exists a possibility that a REST definition exists but there is no corresponding cataloged subroutine or vice versa.

This change uses MVIS admin to additionally validate that the subroutine is cataloged.  If either the REST definition is missing or the subroutine is not cataloged then the server will be deemed invalid and will not allow connections to be opened.  Additionally, when deploying, it will only make the changes that are necessary.  That is, if a REST definition exists but the subroutine is not cataloged then the subroutine will be deployed and cataloged but no changes will be made to the REST definition.  If the subroutine is cataloged but the REST definition does not exist then the REST definition will be created but the subroutine will not be deployed and cataloged.

As a result, this change should improve the verification detection as well only doing the deployment and configuration work necessary to establish the connection.


kthompson23 (approved):
Looks good.

PR #308, opened 2022-11-26
Prevent multiples of same Connection operations

shawnmcknight:
This PR makes two changes to the `Connection` class to prevent consumers from executing the same operations.

1. Connections were being allowed to be opened more than one time, even during the process of opening a connection.  This is an oversight and connections should be prevented from being opened unless they are currently closed.
2. If the server information cache is empty or expired, multiple requests could be made to methods which request server information.  This would result in multiple requests going to the database at the same time unnecessarily since the results of the request are cached.  An exclusive mutex was placed around the acquisition of server information so if additional requests are made while the first is still in flight, the subsequent requests will wait until the first request has finished processing before executing.  This should reduce possible unnecessary database load when the cache expires.


kthompson23 (approved):
Looks great.

PR #357, opened 2023-01-27
Allow specification of a maximum return payload size from the database server

kthompson23:
This PR adds a new configurable option for a maximum allowed return payload size. Either due to bad data or an incorrectly configured query the UniRPC server may return an extremely large payload which can saturate a network or cause MVIS to become unresponsive. If the payload size exceeds the maximum, a DbServerError will be thrown.

I chose a high default value of 100MB but made it a configurable connection and individual request option.

### DIFF
--- a/src/unibasicTemplates/subroutines/internal/setup.njk
+++ b/src/unibasicTemplates/subroutines/internal/setup.njk
@@ -10,8 +10,8 @@ S$MVOM.PROCESS = @TRUE
 gosub processUserDefinedOptions
 
 if udoGetProperty(setupOptions, 'maxReturnPayloadSize', maxReturnPayloadSize, type) then
-  * Value not provided. Default to 1GB
-  maxReturnPayloadSize = 1000000000
+  * Value not provided. Default to 100MB
+  maxReturnPayloadSize = 100000000
 end
 
 parsedSetupOptions<MAX_RETURN_PAYLOAD_SIZE_POS> = maxReturnPayloadSize
--- a/src/unibasicTemplates/subroutines/internal/setup.njk
+++ b/src/unibasicTemplates/subroutines/internal/setup.njk
@@ -1,4 +1,6 @@
-subroutine mvom_setup(setupOptions)
+subroutine mvom_setup(setupOptions, parsedSetupOptions)
+{% include "../../constants/setup.njk" %}
+
 * Perform any needed environment setup
 
 com /S$MVOM/ S$MVOM.PROCESS
@@ -6,6 +8,13 @@ com /S$MVOM/ S$MVOM.USER1, S$MVOM.USER2, S$MVOM.USER3, S$MVOM.USER4, S$MVOM.USER
 S$MVOM.PROCESS = @TRUE
 
 gosub processUserDefinedOptions
+
+if udoGetProperty(setupOptions, 'maxReturnPayloadSize', maxReturnPayloadSize, type) then
+  * Value not provided. Default to 1GB
+  maxReturnPayloadSize = 1000000000
+end
+
+parsedSetupOptions<MAX_RETURN_PAYLOAD_SIZE_POS> = maxReturnPayloadSize
 
 returnFromSub:
 return; * returning to caller
--- a/website/docs/02 - Connection.md
+++ b/website/docs/02 - Connection.md
@@ -27,14 +27,13 @@ Connection.createConnection(mvisUri: string, account: string, options?: CreateCo
 
 #### Options Object Properties
 
-| Property               | Type          | Default         | Description                                                                                                 |
-| ---------------------- | ------------- | --------------- | ----------------------------------------------------------------------------------------------------------- |
-| `logger`               | `Logger`      |                 | An object implementing the [Logger](#logger-interface) interface, used for logging messages emitted by MVOM |
-| `cacheMaxAge`          | `number`      | `3600`          | The maximum age of cached connection information, such as the current database date                         |
-| `timeout`              | `number`      | `0`             | The request timeout in milliseconds (0 to disable)                                                          |
-| `httpAgent`            | `http.Agent`  |                 | An `http.Agent` instance to use with http requests (recommended)                                            |
-| `httpsAgent`           | `https.Agent` |                 | An `https.Agent` instance to use with https requests (recommended)                                          |
-| `maxReturnPayloadSize` | `number`      | `1,000,000,000` | The maximum allowed return payload size specified in bytes. A DbServerError will be thrown                  |
+| Property      | Type          | Default | Description                                                                                                 |
+| ------------- | ------------- | ------- | ----------------------------------------------------------------------------------------------------------- |
+| `logger`      | `Logger`      |         | An object implementing the [Logger](#logger-interface) interface, used for logging messages emitted by MVOM |
+| `cacheMaxAge` | `number`      | `3600`  | The maximum age of cached connection information, such as the current database date                         |
+| `timeout`     | `number`      | `0`     | The request timeout in milliseconds (0 to disable)                                                          |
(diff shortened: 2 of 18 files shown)


src/unibasicTemplates/subroutines/internal/setup.njk
```diff
@@ -7,6 +9,13 @@ S$MVOM.PROCESS = @TRUE
 
 gosub processUserDefinedOptions
 
+if udoGetProperty(setupOptions, 'maxReturnPayloadSize', maxReturnPayloadSize, type) then
```
shawnmcknight, line 12: Is there a concern that we will not have provided a maximum size in the JS code?  It seemed like we should have some guarantees there that the maximum size will have been provided, so this check feels superfluous.

website/docs/02 - Connection.md
```diff
@@ -27,13 +27,14 @@ Connection.createConnection(mvisUri: string, account: string, options?: CreateCo
...
+| `logger`               | `Logger`      |                 | An object implementing the [Logger](#logger-interface) interface, used for logging messages emitted by MVOM |
+| `cacheMaxAge`          | `number`      | `3600`          | The maximum age of cached connection information, such as the current database date                         |
+| `timeout`              | `number`      | `0`             | The request timeout in milliseconds (0 to disable)                                                          |
+| `httpAgent`            | `http.Agent`  |                 | An `http.Agent` instance to use with http requests (recommended)                                            |
+| `httpsAgent`           | `https.Agent` |                 | An `https.Agent` instance to use with https requests (recommended)                                          |
+| `maxReturnPayloadSize` | `number`      | `1,000,000,000` | The maximum allowed return payload size specified in bytes. A DbServerError will be thrown                  |
```
shawnmcknight, line 37: Just for clarification purposes, adding "if this value is exceeded" to to "...will be thrown" sentence would be helpful.

I'm also never quite sure how to format things when multiple sentences are in play.  I feel like a period is unnecessary when its a single sentence, but when there's more than one I feel like periods should end each sentence.

website/docs/04 - Model/03 - Reading By Id.md
```diff
@@ -26,10 +26,11 @@ static findById(id: string, options?: ModelFindByIdOptions): Promise<Model | nul
 
 #### Options Object Properties
 
-| Property      | Type       | Description                                                                                                    |
-| ------------- | ---------- | -------------------------------------------------------------------------------------------------------------- |
-| `projection`  | `string[]` | The [projection](./Advanced%20Topics/model_projection) of properties to return from the database               |
-| `userDefined` | `object`   | The [user defined options](./Advanced%20Topics/model_user_defined_options) to pass to the database subroutines |
+| Property               | Type       | Default         | Description                                                                                                    |
+| ---------------------- | ---------- | --------------- | -------------------------------------------------------------------------------------------------------------- |
+| `maxReturnPayloadSize` | `number`   | `1,000,000,000` | The maximum allowed return payload size specified in bytes. A DbServerError will be thrown                     |
```
shawnmcknight, line 31: I'm not sure if specifying a default for each method is correct.  If not specified, it falls back to the connection's value.  I'm thinking it should either be omitted or some annotation indicating that it will fall back to the connection's value would be appropriate here.

website/docs/04 - Model/05 - Deleting a Model.md
```diff
@@ -26,9 +26,11 @@ static deleteById(id: string, options?: ModelDeleteByIdOptions): Promise<Model |
 
 #### Options Object Properties
 
-| Property      | Type     | Description                                                                                                    |
-| ------------- | -------- | -------------------------------------------------------------------------------------------------------------- |
-| `userDefined` | `object` | The [user defined options](./Advanced%20Topics/model_user_defined_options) to pass to the database subroutines |
+| Property | Type | Default | Description |
+| -------- | ---- | ------- | ----------- |
+
```
shawnmcknight, line 31: Looks like a newline got accidentally added here.

website/docs/04 - Model/06 - Querying/01 - Basics.md
```diff
@@ -26,13 +26,14 @@ static find(queryCondition: Filter, options?: ModelFindOptions): Promise<Model[]
...
-| `sort`        | <code>[string, -1 &#124; 1 ][]</code> | The sort criteria for the query. See [sorting](model_query_sorting).                                            |
-| `projection`  | `string[]`                            | The [projection](../Advanced%20Topics/model_projection) of properties to return from the database               |
-| `userDefined` | `object`                              | The [user defined options](../Advanced%20Topics/model_user_defined_options) to pass to the database subroutines |
+| Property               | Type                                  | Default         | Description                                                                                                     |
+| ---------------------- | ------------------------------------- | --------------- | --------------------------------------------------------------------------------------------------------------- |
+| `maxReturnPayloadSize` | `number`                              | `1,000,000,000` | The maximum allowed return payload size specified in bytes. A DbServerError will be thrown                      |
```
shawnmcknight, line 31: I'd prefer to see this documented lower in the table since it should be a less common option than `skip`, `limit`, etc.  Perhaps placing it just prior to (or just after) `userDefined` in each one of these docs would make a consistent location and keep it below more common options.

My thought here is that the maximum size would most likely be on the connection rather than having overrides.

shawnmcknight (changes requested):
This looks good.  I just had one minor code comment/question and a few tweaks I think should be made to the docs.

With respect to the docs, I'm not sure of the optimal workflow here.
-  If the docs are included with the PR, then updated docs will be deployed before a release occurs.  Therefore features will be documented that are not actually available.
- If the docs are not included with the PR, then they should probably be updated immediately before a release is completed, which complicates the release process.
- If the docs are not completed as part of the release cycle, then there's a risk of them getting stale because its never done.

My thought process on this right now is that the best thing to do would be to create 2 PRs.  The first would be for the actual code change and the other would be for the doc changes.  The code changes could then be merged in first, and then prior to the release being created the docs PR could be merged.  The big risk with this approach is that there might then be multiple PRs open for docs that might be conflicting.

Alternatively, documentation deployment would be adjusted to be a manual rather than an automated process.  Part of the release cycle would be to deploy the latest docs as well.

What are your thoughts?

src/unibasicTemplates/subroutines/internal/setup.njk
kthompson23: Unfortunately it's compile error if `udoGetProperty` isn't in a conditional or doesn't assign its return value. Are you saying just throw the return value away?

shawnmcknight: I see.  So basically you need to do something to get to the value, so it might as well fall back to a default if for some reason there's an error.  We should assume that it should never actually happen that way.  I'm okay with this.

kthompson23:
> 

I think it makes sense to build the docs as part of the release instead of the merge to main.

In this case though I'll create a separate PR for the documentation changes since one of us will need to update the GitHub Action.

kthompson23:
Also, how do you feel about the 1GB default? I set it high because I have no idea what other people do with MVIS but also seems like if it's that high the default value won't catch any issues. Maybe 100MB?

shawnmcknight:
> Also, how do you feel about the 1GB default? I set it high because I have no idea what other people do with MVIS but also seems like if it's that high the default value won't catch any issues. Maybe 100MB?

1GB does seem like a lot considering its solely a parameter for use with MVOM and not for MVIS in general.  I had sort of interpreted that as a "this is basically infinite" default value.  However, if we're going to go with an enormous default value, we might be better off just making it optional.  It's a likely footgun for people if they don't pass it in though, so a sane default is probably good.  I'd be comfortable with 100MB.

kthompson23:
I updated the default size to 100MB and reverted all the documentation changes. I will do a separate documenation PR>

shawnmcknight (approved):
LGTM!

PR #390, opened 2023-03-01
CXM-5221/Pass request id from services

jem41:
This PR adds functionality to pass a request/trace ID through MVOM requests. MVIS 1.3.2 allows a request ID to be specified using the request header `X-MVIS-Trace-Id`. If a request ID is not specified, a random UUID will be generated to be used.

While testing I noticed that MVIS may create separate ROOT actions from a request and will generate a separate X-MVIS-Trace-Id value for those actions. I will update Rocket case 00884089 to ask about these ID values. An example below:

Request ID created by the services for `item.query`: 9619d006-2f05-4b4e-bd25-01d7506242b1

```
[X-MVIS-Trace-Id:SELF: 9619d006-2f05-4b4e-bd25-01d7506242b1] 2023-03-01 16:55:32:850 [qtp1620890840-21035] INFO  com.rs.u2.u2rest.b.c - HTTP request IP address: 10.1.1.201
[X-MVIS-Trace-Id:SELF: 9619d006-2f05-4b4e-bd25-01d7506242b1] 2023-03-01 16:55:32:850 [qtp1620890840-21035] INFO  com.rs.u2.u2rest.b.c - HTTP request user: (anonymous)
[X-MVIS-Trace-Id:SELF: 9619d006-2f05-4b4e-bd25-01d7506242b1] 2023-03-01 16:55:32:887 [qtp1620890840-21035] INFO  com.rs.u2.u2rest.b.c - [dev-107] The tracer data sent to dataserver: 'REST=1'
[X-MVIS-Trace-Id:ROOT: 1367a7ad-0fc2-4b9d-89c8-973d4a50f3f6] 2023-03-01 16:55:33:451 [qtp1620890840-22069] INFO  com.rs.u2.u2rest.b.c - HTTP request IP address: 10.1.2.146
[X-MVIS-Trace-Id:ROOT: 71f426a4-884d-432c-8281-5aec593e6c45] 2023-03-01 16:55:33:451 [qtp1620890840-22068] INFO  com.rs.u2.u2rest.b.c - HTTP request IP address: 10.1.2.146
[X-MVIS-Trace-Id:ROOT: 71f426a4-884d-432c-8281-5aec593e6c45] 2023-03-01 16:55:33:451 [qtp1620890840-22068] INFO  com.rs.u2.u2rest.b.c - HTTP request user: (anonymous)
[X-MVIS-Trace-Id:ROOT: 1367a7ad-0fc2-4b9d-89c8-973d4a50f3f6] 2023-03-01 16:55:33:451 [qtp1620890840-22069] INFO  com.rs.u2.u2rest.b.c - HTTP request user: (anonymous)
````

### DIFF
--- a/src/Connection.ts
+++ b/src/Connection.ts
@@ -289,7 +289,7 @@ class Connection {
 			);
 		}
 
-		const updatedSetupOptions = { ...setupOptions, maxReturnPayloadSize, requestId };
+		const updatedSetupOptions = { ...setupOptions, maxReturnPayloadSize };
 
 		this.logHandler.debug(`executing database subroutine "${subroutineName}"`);
 
--- a/src/Connection.ts
+++ b/src/Connection.ts
@@ -359,7 +359,7 @@ class Connection {
 	}
 
 	/** Get the db server information (date, time, etc.) */
-	private async getDbServerInfo({ requestId }: DbServerInfoOptions): Promise<ServerInfo> {
+	private async getDbServerInfo({ requestId }: DbServerInfoOptions = {}): Promise<ServerInfo> {
 		// set a mutex on acquiring server information so multiple simultaneous requests are not modifying the cache
 		return this.serverInfoMutex.runExclusive(async () => {
 			if (this.dbServerInfo == null || Date.now() > this.dbServerInfo.cacheExpiry) {
--- a/src/Connection.ts
+++ b/src/Connection.ts
@@ -98,7 +98,7 @@ interface ServerInfo {
 }
 
 interface RequestOptions {
-	requestId?: string | null;
+	requestId?: string;
 }
 
 export type OpenOptions = RequestOptions;
--- a/src/Connection.ts
+++ b/src/Connection.ts
@@ -320,24 +320,28 @@ class Connection {
 	}
 
 	/** Get the current ISOCalendarDate from the database */
+	/* istanbul ignore next: Optional parameter is covered by executeDbSubroutine test "should generate requestID if not provided" */
 	public async getDbDate({ requestId }: GetDbDateOptions = {}): Promise<string> {
 		const { timeDrift } = await this.getDbServerInfo({ requestId });
 		return format(addMilliseconds(Date.now(), timeDrift), ISOCalendarDateFormat);
 	}
 
 	/** Get the current ISOCalendarDateTime from the database */
+	/* istanbul ignore next: Optional parameter is covered by executeDbSubroutine test "should generate requestID if not provided" */
 	public async getDbDateTime({ requestId }: GetDbDateTimeOptions = {}): Promise<string> {
 		const { timeDrift } = await this.getDbServerInfo({ requestId });
 		return format(addMilliseconds(Date.now(), timeDrift), ISOCalendarDateTimeFormat);
 	}
 
 	/** Get the current ISOTime from the database */
+	/* istanbul ignore next: Optional parameter is covered by executeDbSubroutine test "should generate requestID if not provided" */
 	public async getDbTime({ requestId }: GetDbTimeOptions = {}): Promise<string> {
 		const { timeDrift } = await this.getDbServerInfo({ requestId });
 		return format(addMilliseconds(Date.now(), timeDrift), ISOTimeFormat);
 	}
 
 	/** Get the multivalue database server limits */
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/Connection.ts [hunk 1]
```diff
@@ -298,26 +320,26 @@ class Connection {
...
-		const { timeDrift } = await this.getDbServerInfo();
+	public async getDbDate({ requestId }: GetDbDateOptions = {}): Promise<string> {
+		const { timeDrift } = await this.getDbServerInfo({ requestId });
 		return format(addMilliseconds(Date.now(), timeDrift), ISOCalendarDateFormat);
 	}
 
 	/** Get the current ISOCalendarDateTime from the database */
-	public async getDbDateTime(): Promise<string> {
-		const { timeDrift } = await this.getDbServerInfo();
+	public async getDbDateTime({ requestId }: GetDbDateTimeOptions = {}): Promise<string> {
```
jem41, line 329: @kthompson23 The optional parameter for these functions is what is failing coverage. I'm not sure what convention would be desired to provide coverage here.

Do I create a separate test to ensure a `requestId` is generated, even though it does not pertain to this function? Do we ignore this line?

src/types/DbSubroutineInput.ts
```diff
@@ -11,6 +11,7 @@ export interface DbSubroutineUserDefinedOptions {
 
 export interface DbSubroutineSetupOptions {
 	userDefined?: DbSubroutineUserDefinedOptions;
+	requestId?: string;
```
shawnmcknight, line 14: At first I was confused as to why this was added to this interface, but after discussing with @kthompson23 it made sense that eventually this `requestId` will make its way to the db server.

I think it might be beneficial to have a comment or something for now indicating that it is for future use by the db server logic as well as for the request header.

src/Connection.ts
```diff
@@ -95,6 +96,22 @@ interface ServerInfo {
 	/** Multivalue database server limits */
 	limits: DbServerLimits;
 }
+
+interface RequestOptions {
+	requestId?: string | null;
```
shawnmcknight, line 101: I don't think we should allow `null` here.  When doing the destructure with defaulting, `null` will not trigger falling back to the default value (i.e. the random UUID) and would instead pass `null` through.  I don't think we want `null` as the header value ever, so the easier approach to resolve that would be to disallow `null` as a valid type here, just leaving it as optional.

src/Connection.ts
shawnmcknight: I'd rather see us reduce the code coverage requirements here than add a pointless test or an istanbul ignore block.

src/Query.ts
```diff
@@ -119,7 +119,7 @@ class Query<TSchema extends GenericObject = GenericObject> {
 
 	/** Execute query */
 	public async exec(options: QueryExecutionOptions = {}): Promise<QueryExecutionResult> {
-		const { maxReturnPayloadSize, userDefined } = options;
+		const { maxReturnPayloadSize, requestId, userDefined } = options;
```
shawnmcknight, line 122: Tests should be augmented for `Query` to cover this new logic.

shawnmcknight (changes requested):
Overall this looks really good.  I had a few requests for change.

jem41:
> Overall this looks really good. I had a few requests for change.

@shawnmcknight Changes made. Do you want me to also lower the coverage thresholds or leave the istanbul ignore blocks for now? If so, should I match the base rules in the nextgen repo or is there another threshold you had in mind?

shawnmcknight:
> > Overall this looks really good. I had a few requests for change.
> 
> @shawnmcknight Changes made. Do you want me to also lower the coverage thresholds or leave the istanbul ignore blocks for now? If so, should I match the base rules in the nextgen repo or is there another threshold you had in mind?
> 
> nextgen config for reference:
> 
> ```
> coverageThreshold: {
> 		'**/*.?([cm])[jt]s?(x)': {
> 			statements: 25,
> 			branches: 25,
> 			functions: 25,
> 			lines: 25,
> 		},
> 	},
> ```

@jem41 How about we let this PR through with the ignores in place.  Then do a quick follow-up PR to make the coverage match the config you indicated here as well as removing all `istanbul ignore` overrides from anywhere in the repo, including the new ones you've added.  That way we take care of this without expanding the scope of this PR.

shawnmcknight (approved):
Looks great!

kthompson23 (approved):
Looks good thank you.

Please do a follow-up PR to update the documentation.

PR #393, opened 2023-03-02
Lower coverage thesholds. Remove istanbull ignores

jem41:
This PR lowers the coverage threshold to 25% for each individual file vs 100% globally. It also removes all `istanbull ignore` comments as they are no longer needed.


shawnmcknight (approved):
LGTM!

PR #397, opened 2023-03-03
Add requestId to common variable

kthompson23:
This PR updates the server-side setup routine to set the `requestId` to a new common variable in our `S$MVOM` common block called `S$MVOM.REQUEST.ID`.

With this change any UniBasic trigger programs that define our common block can utilize the request id for their logging.

I also fixed an uninitialized variable warning in the mvom main program that I introduced when I added support for a max return payload size.

### DIFF
--- a/src/unibasicTemplates/main.njk
+++ b/src/unibasicTemplates/main.njk
@@ -50,6 +50,7 @@ if udoGetProperty(input, 'teardownOptions', teardownOptions, type) then
   go response
 end
 
+parsedSetupOptions = ''
 call mvom_setup(setupOptions, parsedSetupOptions)
 
 subroutineName = 'mvom_':subroutineId
--- a/src/unibasicTemplates/subroutines/internal/setup.njk
+++ b/src/unibasicTemplates/subroutines/internal/setup.njk
@@ -1,13 +1,17 @@
 subroutine mvom_setup(setupOptions, parsedSetupOptions)
+{% include "../../common/mvom.njk" %}
 {% include "../../constants/setup.njk" %}
 
 * Perform any needed environment setup
 
-com /S$MVOM/ S$MVOM.PROCESS
-com /S$MVOM/ S$MVOM.USER1, S$MVOM.USER2, S$MVOM.USER3, S$MVOM.USER4, S$MVOM.USER5
 S$MVOM.PROCESS = @TRUE
 
 gosub processUserDefinedOptions
+
+S$MVOM.REQUEST.ID = ''
+if udoGetProperty(setupOptions, 'requestId', requestId, type) eq 0 then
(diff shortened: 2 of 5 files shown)


src/unibasicTemplates/main.njk
```diff
@@ -50,6 +50,7 @@ if udoGetProperty(input, 'teardownOptions', teardownOptions, type) then
   go response
 end
 
+parsedSetupOptions = ''
```
kthompson23, line 53: Uninitialized variable warning.

src/unibasicTemplates/subroutines/internal/setup.njk
```diff
@@ -1,14 +1,18 @@
...
 * Perform any needed environment setup
 
-com /S$MVOM/ S$MVOM.PROCESS
-com /S$MVOM/ S$MVOM.USER1, S$MVOM.USER2, S$MVOM.USER3, S$MVOM.USER4, S$MVOM.USER5
 S$MVOM.PROCESS = @TRUE
 
 gosub processUserDefinedOptions
 
+S$MVOM.REQUEST.ID = ''
+if udoGetProperty(setupOptions, 'requestId', requestId, type) eq 0 then
```
jem41, line 12: Does `udoGetProperty` return a boolean? Is the logic similar to other languages that 0 is false and otherwise is true?

Just want to understand why the `eq 0` is needed here where I don't see it anywhere else in the repo. Is it for type conversion since requestId is a string?

shawnmcknight: Most (maybe all) of the udo functions return a zero when there is no error, and an error code enumeration otherwise.  The logic elsewhere in this repo is checking to see if there is an error and reacting to that error.  However, this is treating the failure to acquire a requestID as a non-error scenario, so effectively it can be optional.

It's a little more graceful than erroring in this scenario since it doesn't _really_ matter if we don't acquire the request id.

shawnmcknight (approved):
LGTM!

shawnmcknight:
It will probably be good to document this new common variable.  In the advanced topics section there is information about the user defined variables and "detecting mvom in triggers".  Probably would make sense to add a new topic there.

src/unibasicTemplates/subroutines/internal/setup.njk
kthompson23: Also a quirk with UniBasic, `udoGetProperty(setupOptions, 'requestId', requestId, type)` by itself will not compile. You either need to put it in a conditional or assign the return value to a variable.

kthompson23: @jem41 Are you good with this PR?

PR #413, opened 2023-03-14
bump minimum node version to 16.0.0

jem41:
This PR removes support for node version 14, which will be out of support at the end of April. The new minimum version is 16.0.0.

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -3,7 +3,7 @@ module.exports = {
 	env: {
 		build: {
 			ignore: ['**/*.test.ts', '**/__mocks__/**'],
-			presets: [['@babel/env', { targets: { node: '16.19.1' } }], '@babel/typescript'],
+			presets: [['@babel/env', { targets: { node: '16.0.0' } }], '@babel/typescript'],
 		},
 		debug: { sourceMaps: 'inline', retainLines: true },
 	},
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -3,7 +3,7 @@ module.exports = {
 	env: {
 		build: {
 			ignore: ['**/*.test.ts', '**/__mocks__/**'],
-			presets: [['@babel/env', { targets: { node: '14.19.0' } }], '@babel/typescript'],
+			presets: [['@babel/env', { targets: { node: '16.19.1' } }], '@babel/typescript'],
 		},
 		debug: { sourceMaps: 'inline', retainLines: true },
 	},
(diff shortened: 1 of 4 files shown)


kthompson23 (approved):
Looks good, thank you.

Reminder to also do a doc PR.

package.json
```diff
@@ -5,7 +5,7 @@
   "description": "Multivalue Object Mapper",
   "main": "index.js",
   "engines": {
-    "node": ">=14.19"
+    "node": ">=16.19"
```
shawnmcknight, line 8: I'm honestly not sure of best practices here.  Should we support node starting at 16.0 or is it okay to make the range a high patch range?

Presumably we currently work alright with 16.0 and higher since it was in the earlier semver range.  I would think we'd want to keep this band wider unless we know of something that will not work with an earlier version.

.babelrc.js
```diff
@@ -3,7 +3,7 @@ module.exports = {
 	env: {
 		build: {
 			ignore: ['**/*.test.ts', '**/__mocks__/**'],
-			presets: [['@babel/env', { targets: { node: '14.19.0' } }], '@babel/typescript'],
```
shawnmcknight, line 6: When targeting, we'll want to match whatever the lowest end of the range would be.  If we change the range then that would match that (i.e. `16.0.0`).  If we leave this higher range then it would be `16.19.0`).

shawnmcknight (commented):
The changes look good, but I am wondering if we should keep the range a bit wider to include all of 16.

What do you think @kthompson23 ?

kthompson23:
> 



> The changes look good, but I am wondering if we should keep the range a bit wider to include all of 16.
> 
> What do you think @kthompson23 ?

@shawnmcknight @jem41 Yea thinking about this more I don't think I've ever seen a minimum node version pinned to the minor version. I think we started pinning to an exact version of node 12 or 14 because we wanted to use async local storage in the services.

It would be more user friendly if we set the minimum requirement as node 16.

jem41:
Minimum constraint updated to base 16 and babel target to 16.0.0.

shawnmcknight (changes requested):
Changes look good -- can you update the PR title and description to match the node version?

Also, since this is OSS, can you remove the internal task ID from the description?

kthompson23 (approved):
Thank you.

jem41:
> Changes look good -- can you update the PR title and description to match the node version?
> 
> Also, since this is OSS, can you remove the internal task ID from the description?

Fixed.

shawnmcknight (approved):
LGTM!

kthompson23 (approved):

PR #417, opened 2023-03-15
Cxm 5292/connection documentation

jem41:
This PR documents the changes made in PRs https://github.com/STORIS/mvom/pull/186 and https://github.com/STORIS/mvom/pull/196 to automatically deploy the REST definition and subroutine code if not present.


website/docs/01 - Introduction/03 - Setup and Configuration.md
```diff
@@ -10,352 +10,7 @@ In order to communicate to the MultiValue database, the following are required:
...
-* note that this routine intentionally does not react to udo errors because it could
-* potentially result in an unending sequence of errors
-
-x = udoSetProperty(output, 'errorCode', errorCode)
-
-return
-```
-
-Upon clicking "OK", MVIS Admin will allow for the source code to be deployed.
+MVOM will automatically create a REST subroutine definition and deploy the subroutine to the database server if not present. The created subroutine definition is named `mvom_main` and is suffixed by an 8 character hash of the bundled source code. A new version of MVOM will result in a new hash and associated subroutine deployment.
```
shawnmcknight, line 16: "..._will_ result..." is not necessarily accurate.  Only if the bundle changes will a new subroutine hash be generated.  "_may_ result"  might be a better phrasing here.

website/docs/02 - Connection.md
```diff
@@ -89,59 +92,6 @@ const makeConnection = async (): Connection => {
 export default makeConnection;
 ```
 
-## Deploying MVOM database server features
```
shawnmcknight, line 92: While there is no longer a `deployFeatures` method on the `Connection` class, there is still a `deploy` method on the class and deployment is still something that the consumer of the `Connection` class must manually do.  If the DB server deployment isn't valid, the `open` method will throw and a connection won't be opened.

Basically most of this section is still pertinent and just needs some tweaks.

shawnmcknight (changes requested):
Finished a review.  The deployment section is still necessary in the `Connection` class documentation.

jem41:
Restored the deploy section of `Connection.md` and updated to conform with the new method signatures. Also changed the verbiage from _will_ to _may_ in `Setup and Configuration.md`

website/docs/02 - Connection.md
```diff
@@ -96,7 +99,7 @@ MVOM requires a number of database server subroutines (referred to by MVOM as _s
 ### Syntax
```
shawnmcknight, line 99: I can't tag the appropriate spot in the file, but I think the description text above is a little misleading now since we only deploy a single subroutine.  I think it just needs some slight tweaking to refer to the `mvom_main` subroutine.

shawnmcknight (approved):
LGTM!

PR #463, opened 2023-05-02
Standardize logging levels

shawnmcknight:
This PR adjusts the logging levels being used by the library, removing `verbose` and `silly` and adding `fatal` and `trace`.


kthompson23 (approved):
Makes sense. Looks good, thanks.

PR #465, opened 2023-05-03
Remove use of @storis/prettier-config and internalize configuration

shawnmcknight:
This PR removes the use of `@storis/prettier-config` to set the configuration for prettier as that package is no longer being maintained.  A `.prettierrc.json` file has been added to internalize the settings for prettier.


kthompson23 (approved):
Looks good.

PR #466, opened 2023-05-04
Update dependencies

shawnmcknight:
This PR breaks the logjam of dependency updates that renovate could not push through due to peer dependency violations.  Every outstanding dependency that is not current has been updated.

There were several breaking changes in the `@storis/eslint-config` and `@storis/tsconfig` packages which needed to be resolved by changing the config file structures.  Additionally, some lint errors cropped up which needed to be fixed.


kthompson23 (approved):
Looks good, thank you.

PR #467, opened 2023-05-04
CXM-5384 Fix fk validation with arrays of file names

rmattos500:
It was recently discovered that even though we can specify an array of file names in the `foreignKey` property of our mvom schemas to validate the foreign key against, it is only ever validating against the first file. This PR changes the file name/entity name separator character from a `,` to a `#` since when an array of file names gets transformed to a string it inserts commas between each of them. Thus, when performing splitting logic later on in the process flow, it was not working correctly. After making this change the file names, were correctly split from the entity name, but additional splitting was needed to split each of the file names back into an array to pass to the u2 subroutine.

### DIFF
--- a/src/Document.ts
+++ b/src/Document.ts
@@ -12,7 +12,7 @@ export interface DocumentConstructorOptions {
 }
 
 export interface BuildForeignKeyDefinitionsResult {
-	filename: string;
+	filename: string[];
 	entityName: string;
 	entityIds: string[];
 }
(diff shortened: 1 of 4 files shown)


src/Document.ts
```diff
@@ -165,7 +165,10 @@ class Document {
 		return Array.from(definitionMap).reduce<BuildForeignKeyDefinitionsResult[]>(
 			(acc, [key, value]) => {
 				const keyParts = key.split(separator);
-				const filename = keyParts[0];
+				const fileName = keyParts[0];
+				// If an array of filenames was provided, when we transformed the array into a string above, commas
+				// would have been inserted between each filename. Split the string back into an array.
+				const filename = fileName.split(',');
 				// Just incase the entity name included a comma, rejoin
```
kthompson23, line 172:
```suggestion
				// Just incase the entity name included a pound sign, rejoin
```

kthompson23 (commented):
Looks good other than the one comment now being out dated. If it looks good, just accept my suggestion.

rmattos500:
That's really cool that it allows you to just accept a suggestion like that. I accepted it, thanks!

PR #639, opened 2024-04-10
Improve observability in validation routines

philfuster:
This PR aims to increase the observability into the validation routines within MVOM. As it stands, when a validation error is encountered, you know the key path of the property trying to be validated. You don't see the value, nor do you know the keyPath of the key the error actually happened on when dealing with embedded documents or nested arrays.
To address that, this PR adds `keyPath` to the `validate` and `validateType` methods and also will log the value.

Logging `keyPath` on a scalar attribute can feel redundant because you would see the `keyPath` twice. Once in the error set as the key the error was thrown for, but also in the individual error that was thrown.
The real benefit comes with nested documents because this additional logging should show they keyPath the error was actually thrown on in nested documents.


philfuster:
@shawnmcknight I'm going to start writing tests for this now, but I wanted to get some eyes on it in case you have other ideas or concerns. 
schema
record trying to create with missing required property and invalid prop3 value.
resulting errors logged

Attached are screen shots of the schema I used to test, the record I was trying to create, and the error message that was logged.

I think this'll be helpful in some of the errors we were seeing thrown, because it was a `details` property but we couldn't see what property that was required was actually missing. In the logged message attached you see that now you'd see that it was the parent property `details` that had the error and the `prop1` property that was missing.

If you have any additional thoughts let me know. I'm going start writing some tests.

philfuster:
> @shawnmcknight I'm going to start writing tests for this now, but I wanted to get some eyes on it in case you have other ideas or concerns. schema record trying to create with missing required property and invalid prop3 value. resulting errors logged
> 
> Attached are screen shots of the schema I used to test, the record I was trying to create, and the error message that was logged.
> 
> I think this'll be helpful in some of the errors we were seeing thrown, because it was a `details` property but we couldn't see what property that was required was actually missing. In the logged message attached you see that now you'd see that it was the parent property `details` that had the error and the `prop1` property that was missing.
> 
> If you have any additional thoughts let me know. I'm going start writing some tests.

Thinking about it some more. Maybe instead of just a raw string for the error message it should e more structural and be an object with a `message`, `keyPath`/`nestedKeyPath` property and `value` property.

shawnmcknight:
> > @shawnmcknight I'm going to start writing tests for this now, but I wanted to get some eyes on it in case you have other ideas or concerns. schema record trying to create with missing required property and invalid prop3 value. resulting errors logged
> > Attached are screen shots of the schema I used to test, the record I was trying to create, and the error message that was logged.
> > I think this'll be helpful in some of the errors we were seeing thrown, because it was a `details` property but we couldn't see what property that was required was actually missing. In the logged message attached you see that now you'd see that it was the parent property `details` that had the error and the `prop1` property that was missing.
> > If you have any additional thoughts let me know. I'm going start writing some tests.
> 
> Thinking about it some more. Maybe instead of just a raw string for the error message it should e more structural and be an object with a `message`, `keyPath`/`nestedKeyPath` property and `value` property.

So, this is really about trying to get a better path logged when a validation error occurs on a document array or an embedded document, right?

In those scenarios, the `schemaType` validations run the document validator. That validates each property independently and then creates a `Map` with a key of the `keyPath` and a value that is the array of string errors. The `DocumentArrayType` and `EmbeddedType` validators then iterate the values of the `Map` and returns the list of all errors in the subdocument.

So at the time we run those embedded validations, the inner `keyPath` is available to the validation function in the returned errors list. As a high level theory, what if the `BaseSchemaType.validate` function was changed so its signature allowed it to return a `string[] | Map<string, string[]>`? In that case, the validators could return the result they get back instead of transforming it into a `string[]`. The `Document.validate` could then inspect the returned value to see if it's an array or a map. If its an array, it would do what it does today -- add the `keyPath` to its own map and return the list of errors. It it's a map, it could iterate the map it received and then augment its own map with a `.` concatenation of its current `keyPath` and the returned `keyPath`. Something like:
```ts
            if (Array.isArray(errors) && errors.length > 0) {
              documentErrors.set(keyPath, errors);
            }

            if (errors instanceof Map && errors.size > 0) {
              errors.forEach((error, key) => {
                documentErrors.set(`${keyPath}.${key}`, error);
              });
            }
```

That should in theory roll out the errors and the keyPaths without having to funnel them down to the individual validators.

philfuster:
> > > @shawnmcknight I'm going to start writing tests for this now, but I wanted to get some eyes on it in case you have other ideas or concerns. schema record trying to create with missing required property and invalid prop3 value. resulting errors logged
> > > Attached are screen shots of the schema I used to test, the record I was trying to create, and the error message that was logged.
> > > I think this'll be helpful in some of the errors we were seeing thrown, because it was a `details` property but we couldn't see what property that was required was actually missing. In the logged message attached you see that now you'd see that it was the parent property `details` that had the error and the `prop1` property that was missing.
> > > If you have any additional thoughts let me know. I'm going start writing some tests.
> > 
> > 
> > Thinking about it some more. Maybe instead of just a raw string for the error message it should e more structural and be an object with a `message`, `keyPath`/`nestedKeyPath` property and `value` property.
> 
> So, this is really about trying to get a better path logged when a validation error occurs on a document array or an embedded document, right?
> 
> In those scenarios, the `schemaType` validations run the document validator. That validates each property independently and then creates a `Map` with a key of the `keyPath` and a value that is the array of string errors. The `DocumentArrayType` and `EmbeddedType` validators then iterate the values of the `Map` and returns the list of all errors in the subdocument.
> 
> So at the time we run those embedded validations, the inner `keyPath` is available to the validation function in the returned errors list. As a high level theory, what if the `BaseSchemaType.validate` function was changed so its signature allowed it to return a `string[] | Map<string, string[]>`? In that case, the validators could return the result they get back instead of transforming it into a `string[]`. The `Document.validate` could then inspect the returned value to see if it's an array or a map. If its an array, it would do what it does today -- add the `keyPath` to its own map and return the list of errors. It it's a map, it could iterate the map it received and then augment its own map with a `.` concatenation of its current `keyPath` and the returned `keyPath`. Something like:
> 
> ```ts
>             if (Array.isArray(errors) && errors.length > 0) {
>               documentErrors.set(keyPath, errors);
>             }
> 
>             if (errors instanceof Map && errors.size > 0) {
>               errors.forEach((error, key) => {
>                 documentErrors.set(`${keyPath}.${key}`, error);
>               });
>             }
> ```
> 
> That should in theory roll out the errors and the keyPaths without having to funnel them down to the individual validators.

Thank you, Shawn. I will give this a try today.

philfuster:
Closing as I test things out so the pipeline doesn't run every time I push up.

philfuster:
@shawnmcknight I followed your advice and this is the result. I'm going to start writing tests. Attached is the test program I used to test this out. What I noticed was the `embeddedObject` was actually fine. In the current implementation it does concatenate the `keyProp` of the parent and the `keyProp` of the embedded object. I found the `objectArray` and `arrayType` were lacking some context. I added logging of the index for the `arrayType` validation, which I think will be useful.

I am going to start on tests again, but wanted to get your feedback. Thank you.

PR #642, opened 2024-04-16
Improve observability in validation routines

philfuster:
This PR improves the error message reporting during validation to include the expanded key path for imbedded document/document arrays. Scalar arrays will also report the index the error relates to.


src/schemaType/BaseSchemaType.ts
```diff
@@ -64,7 +64,10 @@ abstract class BaseSchemaType {
 	public abstract set(record: MvRecord, value: unknown): MvRecord;
 
 	/** Validate value */
-	public abstract validate(value: unknown, document: Document): Promise<string[]>;
+	public abstract validate(
+		value: unknown,
+		document: Document,
+	): Promise<string[] | Map<string, string | string[]> | Map<string, string | string[]>[]>;
```
shawnmcknight, line 70: Can this return type be simplified? I'd rather see this be `string[] | Map<string, string[]>`. There's a lot of potential combinations here and it complicates the analysis of the response.

src/Document.ts
```diff
@@ -199,9 +200,16 @@ class Document {
...
+						if (errors instanceof Map || this.#isArrayOfMaps(errors)) {
+							ensureArray(errors).forEach((errorMap) => {
+								errorMap.forEach((error, key) => {
+									documentErrors.set(`${keyPath}.${key}`, error);
+								});
+							});
+						}
 					} catch (err) {
 						// an error was thrown - return the message from that error in the documentErrors list
 						documentErrors.set(keyPath, err.message);
```
shawnmcknight, line 215: I think the `documentErrors` type should be narrowed to `Map<string, string[]`>. It appears to me that this is the only scenario where a `string` is the `Map` value. If this set the value as `[err.message`] instead I think this type can be simplified a lot.

src/schemaType/ArrayType.ts
```diff
@@ -37,7 +37,11 @@ class ArrayType extends BaseScalarArrayType {
 	public async validate(value: unknown, document: Document): Promise<string[]> {
 		return (
 			await Promise.all(
-				ensureArray(value).map((arrayItem) => this.valueSchemaType.validate(arrayItem, document)),
+				ensureArray(value).map(async (arrayItem, index) => {
+					const result = await this.valueSchemaType.validate(arrayItem, document);
+
+					return result.map((message) => `index ${index}: ${message}`);
```
shawnmcknight, line 43: Rather than making a special message here, couldn't this return the array index as the map key representing the keyPath?

src/schemaType/DocumentArrayType.ts
```diff
@@ -60,21 +60,18 @@ class DocumentArrayType extends BaseSchemaType {
...
-		).flat(2);
+		const result = await Promise.all(
+			documentList.map(async (document) => {
+				const documentErrors = await document.validate();
+
+				return documentErrors;
+			}),
+		);
+
+		return result;
```
shawnmcknight, line 74: With the other recommended change to use the array index as a keypath, the same thing could be done here. Instead of returning an array of `Map` instances, construct a singular `Map` that takes the returned `Map`'s keypath and prefixes it with the array index.

src/Document.ts
shawnmcknight: The `DataValidationError` type can also have its type narrowed in the same way.

shawnmcknight (changes requested):
Overall looks good, but I think there is room for simplification on how the validation output is structured. Doing so will make this a lot easier to work with and structure.

src/schemaType/ArrayType.ts
philfuster: Good idea.

philfuster:
@shawnmcknight I believe I have made all requested changes. This is ready for another look. Thank you.

src/schemaType/DocumentArrayType.ts
```diff
@@ -60,21 +60,26 @@ class DocumentArrayType extends BaseSchemaType {
 	}
 
 	/** Validate the document array */
-	public async validate(documentList: Document[]): Promise<string[]> {
+	public async validate(documentList: Document[]): Promise<Map<string, string[]>> {
 		// combining all the validation into one array of promise.all
 		// - validation against the documents in the array will return a single object with 0 to n keys - only those with keys indicate errors;
```
shawnmcknight, line 65: This comment seems incorrect. The validation returns a `Map`, correct?

src/schemaType/DocumentArrayType.ts
```diff
@@ -60,21 +60,26 @@ class DocumentArrayType extends BaseSchemaType {
...
-			)
-		).flat(2);
+		const errorsMap = new Map<string, string[]>();
+		await Promise.all(
+			documentList.map(async (document, index) => {
+				const documentErrors = await document.validate();
+
+				const indexString = String(index);
+				documentErrors.forEach((messages, key) => {
+					const errors = errorsMap.get(key) ?? [];
```
shawnmcknight, line 73: This doesn't seem correct. The `errorsMap` is keyed by the combined keypath (as seen in the `set` call below), but this is getting by the keypath returned from the validation routine.

src/schemaType/__tests__/ArrayType.test.ts
```diff
@@ -224,6 +227,6 @@ describe('validate', () => {
 		const value = [1.23, 4.56, 7.89];
 
 		const validationResult = await arrayType.validate(value, documentMock);
-		expect(validationResult).toHaveLength(0);
+		expect(validationResult).toEqual(new Map<string, string[]>());
```
shawnmcknight, line 230: This can probably just be `expect(validationResult.size).toBe(0)` and accomplish the same goal without constructing an empty map.

src/schemaType/__tests__/DocumentArrayType.test.ts
```diff
@@ -285,7 +290,7 @@ describe('validate', () => {
 		value2.prop2 = 4.56;
 
 		const validationResults = await documentArrayType.validate([value1, value2]);
-		expect(validationResults).toEqual([]);
+		expect(validationResults).toEqual(new Map());
```
shawnmcknight, line 293: Same note here with checking the map's size.

src/schemaType/__tests__/EmbeddedType.test.ts
```diff
@@ -80,24 +80,26 @@ describe('validate', () => {
...
+	test('should return a map of the errors encountered when validating the subdocument', async () => {
 		const originalRecord: MvRecord = ['unrelated'];
 		const value = Document.createSubdocumentFromRecord(valueSchema, originalRecord);
 
 		const validationResults = await embeddedType.validate(value);
-		expect(validationResults).toHaveLength(2);
-		validationResults.forEach((result) => {
-			expect(result).toBe('Property is required');
-		});
+		expect(validationResults).toEqual(
```
shawnmcknight, line 88: You're using the `const expected...` pattern elsewhere. It would be good to be consistent here.

src/schemaType/__tests__/EmbeddedType.test.ts
```diff
@@ -80,24 +80,26 @@ describe('validate', () => {
...
-	test('should return an empty array if no errors are encountered when validating the subdocument', async () => {
+	test('should return an empty map if no errors are encountered when validating the subdocument', async () => {
 		const originalRecord: MvRecord = ['unrelated'];
 		const value = Document.createSubdocumentFromRecord(valueSchema, originalRecord);
 		value.prop1 = 'foo';
 		value.prop2 = 12.34;
 
 		const validationResults = await embeddedType.validate(value);
-		expect(validationResults).toEqual([]);
+		expect(validationResults).toEqual(new Map());
```
shawnmcknight, line 103: Same note about checking map size.

src/Document.ts
```diff
@@ -199,12 +199,18 @@ class Document {
 						setIn(this, keyPath, value);
 
 						const errors = await schemaType.validate(value, this);
-						if (errors.length > 0) {
+						if (errors instanceof Map) {
+							errors.forEach((error, key) => {
```
shawnmcknight, line 203: `error` is a bit confusing of a variable name here since its an array. I would suggest using a name that implies that.

src/__tests__/Document.test.ts
```diff
@@ -856,7 +856,46 @@ describe('validate', () => {
 			});
 			const document = new DocumentSubclass(schema, { data: { prop1: 'foo' } });
 
-			const expected = new Map([['prop1', 'Test error message']]);
+			const expected = new Map([['prop1', ['Test error message']]]);
+			expect(await document.validate()).toEqual(expected);
+		});
+
+		test('should return error at property & unravel nested errors if schemaType validation fails', async () => {
```
shawnmcknight, line 863: This test is missing coverage of setting a schema to a property (effectively as an embedded object) and setting a schema as an object-array definition.

I think that might be what you're trying to do with the `nestedSchema` and `subDocumentArraySchema` but those are just objects and not instances of `Schema`.

It looks like you've covered the non-schema version of an embedded object with the `nestedObject` structure, so switching `nestedSchema` to be a `Schema` instance will cover both directly embedding and having a schema. You don't have two versions of a subdocument array, so you'd need to exercise both.

I think you should be careful with the wording as you've got a `nestedObject` property and then an `embeddedType` which uses something called `nestedSchema`. That gets a bit confusing.

src/__tests__/Document.test.ts
```diff
@@ -856,7 +856,46 @@ describe('validate', () => {
...
+				},
+				subdocumentArray: [subDocumentArraySchema],
+				embeddedType: nestedSchema,
+			};
+			const schema = new Schema(definition);
+			const document = new DocumentSubclass(schema, {
+				record: ['55', null, ['foo', 'bar'], [null], 'bing', null, [[null, '5']], [['5', null]]],
+			});
+
+			const expected = new Map([
```
shawnmcknight, line 891: I think this would be a better exercise if you had "good" and "bad" variants of each type. For instance, there's no exercising of the regular array here because its passing. This test therefore doesn't confirm that arrays are working as expected.

So I think you can make have each data structure represented 2x, one that should pass and one that should fail. You could prefix each with `passing` and `failing` or something like that.

shawnmcknight (changes requested):
Looks good, I think this is really close. I had a few comments for review -- mostly about tests but had some more specific concerns about the handling of document arrays.

src/__tests__/Document.test.ts
```diff
@@ -856,7 +856,133 @@ describe('validate', () => {
 			});
 			const document = new DocumentSubclass(schema, { data: { prop1: 'foo' } });
 
-			const expected = new Map([['prop1', 'Test error message']]);
+			const expected = new Map([['prop1', ['Test error message']]]);
+			expect(await document.validate()).toEqual(expected);
+		});
+
+		test('return error at property & unravel nested errors if schemaType validation fails', async () => {
+			const passingEmbeddedObjectSchema: SchemaDefinition = {
```
shawnmcknight, line 864: These variables which are merely schema definitions are confusing when named `Schema` as they imply that they are schemas. I don't think there's much value in defining them as separate variables -- generally we see Schemas with embedded structures as being defined inline with the `Schema` constructor rather than as separate variables. By inlining it you won't have to worry about confusing variable names.

src/__tests__/Document.test.ts
```diff
@@ -856,7 +856,133 @@ describe('validate', () => {
...
+			const passingNestedSchema = new Schema({
+				prop1: { type: 'string', path: '15', required: true },
+				prop2: { type: 'number', path: '16', dbDecimals: 2, required: true },
+			});
+			const failingNestedSchema = new Schema({
+				prop1: [{ type: 'string', path: '17', required: true }],
+				prop2: { type: 'number', path: '18', dbDecimals: 2, required: true },
+			});
+
+			const passingNestedSchemaDocumentArraySchema = new Schema({
```
shawnmcknight, line 891: These are being referred to as "nested", but I wouldn't think of them as such. You can eliminate the `NestedSchema` part of the name and make it easier to read and understand.

src/__tests__/Document.test.ts
```diff
@@ -856,7 +856,133 @@ describe('validate', () => {
...
+			const definition: SchemaDefinition = {
+				passingProp: { type: 'string', path: '1', required: true },
+				failingProp: { type: 'number', path: '2', dbDecimals: 2, required: true },
+				passingArray: [{ type: 'string', path: '3', required: true }],
+				failingArray: [{ type: 'string', path: '4', required: true }],
+				passingNestedArray: [[{ type: 'string', path: '5', required: true }]],
+				failingNestedArray: [[{ type: 'string', path: '6', required: true }]],
+				passingEmbeddedObject: passingEmbeddedObjectSchema,
+				failingEmbeddedObject: failingEmbeddedObjectSchema,
+				passingEmbeddedObjectDocumentArray: [passingEmbeddedObjectDocumentArraySchema],
```
shawnmcknight, line 908: Similarly here, I wouldn't refer to this as an "embedded object".

shawnmcknight (changes requested):
Looks really good. I had a concern that we're doing work unnecessarily in the document array and suggest some clarification on variable names on the document test.

src/schemaType/DocumentArrayType.ts
```diff
@@ -60,21 +60,26 @@ class DocumentArrayType extends BaseSchemaType {
...
+		// Create a map to store the errors. The key will be the index of the document in the list prefixed to the key path of the error
+		const errorsMap = new Map<string, string[]>();
+		await Promise.all(
+			documentList.map(async (document, index) => {
+				const documentErrors = await document.validate();
+
+				const indexString = String(index);
+				documentErrors.forEach((messages, keyPath) => {
+					const errorsMapKey = `${indexString}.${keyPath}`;
+					const errors = errorsMap.get(errorsMapKey) ?? [];
```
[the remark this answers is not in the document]
philfuster: I would think there is no scenario where the map already contains errors for that key path, as you said, it's validating a separate doc at each position in the array, so the array index would be the difference in the built map keys.

I'll make the change and directly put the `messages` at the built keypath in `errorsMap` when there's > 0 messages.

src/__tests__/Document.test.ts
philfuster: Making that change. Leaving `Object` so it can differentiate from the use of `new Schema`.

src/schemaType/DocumentArrayType.ts
```diff
@@ -60,21 +60,24 @@ class DocumentArrayType extends BaseSchemaType {
...
+	public async validate(documentList: Document[]): Promise<Map<string, string[]>> {
+		// Create a map to store the errors. The key will be the index of the document in the list prefixed to the key path of the error
+		const errorsMap = new Map<string, string[]>();
+		await Promise.all(
+			documentList.map(async (document, index) => {
+				const documentErrors = await document.validate();
+
+				const indexString = String(index);
+				documentErrors.forEach((messages, keyPath) => {
+					const errorsMapKey = `${indexString}.${keyPath}`;
```
shawnmcknight, line 72: The setting of this variable can be moved inside the `if` block below since its only used in that scenario.

src/schemaType/ArrayType.ts
```diff
@@ -34,12 +34,22 @@ class ArrayType extends BaseScalarArrayType {
...
-				ensureArray(value).map((arrayItem) => this.valueSchemaType.validate(arrayItem, document)),
-			)
-		).flat();
+	public async validate(value: unknown, document: Document): Promise<Map<string, string[]>> {
+		const errorsMap = new Map<string, string[]>();
+		await Promise.all(
+			ensureArray(value).map(async (arrayItem, index) => {
+				const result = await this.valueSchemaType.validate(arrayItem, document);
+
+				return result.forEach((message) => {
```
shawnmcknight, line 43: I didn't catch this on the last pass, but I think this logic is more complicated that it needs to be. I think this could check to see if `result.length > 0` and set the map value to whatever is in `result` if it is.

shawnmcknight (changes requested):
Tests are looking good now. I think I found two more areas where the syntax can be improved however.

src/schemaType/DocumentArrayType.ts
philfuster: Good idea. done.

src/schemaType/ArrayType.ts
philfuster: good idea. i will make this change.

src/schemaType/NestedArrayType.ts
```diff
@@ -41,14 +41,24 @@ class NestedArrayType extends BaseScalarArrayType {
...
-			)
-		).flat();
+	public async validate(value: unknown, document: Document): Promise<Map<string, string[]>> {
+		const errorsMap = new Map<string, string[]>();
+		await Promise.all(
+			ensureArray(value).map(async (arrayItem, index) => {
+				await Promise.all(
+					ensureArray(arrayItem).map(async (nestedArrayItem, nestedIndex) => {
+						const result = await this.valueSchemaType.validate(nestedArrayItem, document);
+						const key = `${index}.${nestedIndex}`;
```
shawnmcknight, line 51: The same optimizations can be made here.

src/schemaType/EmbeddedType.ts
```diff
@@ -48,15 +47,10 @@ class EmbeddedType extends BaseSchemaType {
...
+	public async validate(document: Document): Promise<Map<string, string[]>> {
 		// - validation against the embedded document will return a single object with 0 to n keys - only those with keys indicate errors;
-		// - iterate the returned object and return the messages from each
-		// - flatten the final results
 		const documentErrors = await document.validate();
-
-		return Array.from(documentErrors.values())
-			.map((err) => ensureArray(err))
-			.flat();
+		return documentErrors;
```
shawnmcknight, line 53: This method can be changed to no longer have async/await and just return the result

shawnmcknight (changes requested):
I noticed that the nested array can have the same optimizations done and also there is an opportunity to eliminate an async/await.

philfuster:
@shawnmcknight Thank you for the reviews. Ready for another look.

shawnmcknight (approved):
Looks good, thanks!

PR #653, opened 2024-04-24
Fix Projection

philfuster:
This PR fixes the format_document u2 subroutine's handling of the `projection` option.
Before this fix, when formatting the output document it was not using the passed in `projection`. It was using the loop index. This PR ensures the provided projection positions get used when formatting the output document.

### DIFF
--- a/src/__tests__/compileModel.test.ts
+++ b/src/__tests__/compileModel.test.ts
@@ -210,6 +210,42 @@ describe('find', () => {
 			{ maxReturnPayloadSize, userDefined, requestId },
 		);
 	});
+
+	test('projection test', async () => {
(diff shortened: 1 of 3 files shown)


shawnmcknight:
@mmw-storis I believe you did the original implementation of projection in mvom. Can you review this?

mmw-storis (approved):
@philfuster I think it worked before because I added this projection to get the first attribute only and array position and actual attribute position are both 1. Once you try to get other attributes, the code will fail because it used the array position instead of the actual attribute position. 
Your fix looks good to me. Thanks!!

shawnmcknight (approved):
LGTM!

PR #697, opened 2024-06-13
Add output object type inference based upon Schema definition

shawnmcknight:
This PR enhances the capabilities of MVOM for type inference based on the schema definition. These changes allow, given a schema, to infer the output types of either a `Document` or a `Model` instance.

The following is a summary of the changes:

### `Schema` class
The `Schema` class has been changed so that there is now a generic `TSchemaDefinition` on the class which will be inferred based on the provided `definition` parameter to the `Schema` constructor. This generic is the basis for all other changes included in this PR.

Several new utility types have been added to the `schema.ts` module to allow for output object type inference:

The exported `InferDocumentObject` type accepts a `Schema` as a generic and will output a type which aligns with the structure of a `Document` that was instantiated based upon that schema's definition.

The exported `InferModelObject` type accepts a `Schema` as a generic and will output a type which aligns with the structure of a `Model` that was instantiated based upon that schema's definition. The primary difference between this type and the `InferDocumentObject` type is that a `Model` instance will include the `_id` and `__v` properties, so this is effectively an extension of the `InferDocumentObject` type.

An internal `InferSchemaType` type provides most of the work regarding the inference. It will accept a schema's type definition (that which is assigned to a schema's property value) and recursively process it to determine the output type of a property in the schema's definition. It will handle each scalar type as well as embedded definitions, embedded schemas, scalar arrays, nested scalar arrays, and document arrays.

An internal `InferStringType` type provides additional utility for string schema type definitions. Since those type definitions may have an enumeration constraint, this type will check if there is a defined enumeration in the definition. If there is, and it is a readonly array, then the output of the string type will be a union of the enumerated strings.

An internal `InferRequiredType` type will detect whether a scalar type is required or not. If it is not required then the resulting output will be unioned with `null`.

#### Usage
Consumers of MVOM can use the `InferModelObject` and `InferDocumentObject` types to generate a type which will provide the shape of the output for a `Model` or `Document` as follows:

```ts
import { Schema } from 'mvom';
import type { InferDocumentObject, InferModelObject } from 'mvom';

const schema = new Schema({ stringProp: { type: 'string', path: '1' } });

type DocumentOutput = InferDocumentObject<typeof schema>; // { stringProp: string | null }
type ModelOutput = InferModelObject<typeof schema>; // { _id: string; __v: string; stringProp: string | null }
```

#### Testing of the utility types
The `Schema.test.ts` suite was updated to have several new tests to confirm that the utility types emit the expected output type. A new utility type named `Assert` was created which will compare two types and return `true` if they match and an error output if they do not. This `Assert` type was used to compare various schemas to their expected output.

This test suite can also provide a lot of insight into the expected output types for various schema definitions.

Note: A failed type assertion would not fail any unit tests, but it would trigger typechecking errors. They have the same end result which is that the CI suite would fail.

### compileModel.ts & dynamic `Model` class
The compileModel function has been augmented to take advantage of the `InferModelObject` inference. The `schema` provided to the function will have its output inferred. This output is used to form a new `ModelCompositeValue` type which is the intersection of the `Model` instance and the inferred output. The various static and instance methods on the `Model` class which  return instances of the `Model` will now be of this composite object. Effectively, those methods will now return a type which has all of the properties strongly typed as defined by the schema.

Additionally, when instantiating a new `Model`, the `data` property supplied to the constructor must now comply with the inferred object shape. That is, there is type safety applied to the data that would construct a `Model` instance.

### `Document` class
Similar to the changes made to the `Model` class, the `Document` class now outputs a `DocumentCompositeValue` from the static methods that can instantiate a new `Document` instance. This composite type will be the intersection of the `Document` instance and the inferred output object.

### Schema type validator changes
Prior to this PR, the validators for a schema type would accept not only the value being validated, but also the `Document` instance they were being validated for. The passing of the `Document` instance was never used by any validators. Because the `Document` is now generic, this complicated the validation as all schema types would have needed to be provided the type of the `Schema` which they were a member of. Since this validation was never used anywhere, it was far simpler to simply remove the document parameter from the validators.

### Query class changes
The `Query` class accept a `Model` constructor previous to this PR. However, TypeScript was complaining about this constructor due to the new composite output type of the `Model` methods. Instead, the `Query` class was modified to individually accept as parameters the things it was using from the `Model` constructor -- the connection, schema, and file. The `Model` method were adjusted to provide this information directly instead of providing its own constructor. This should have no impact to anything with the `Query` class as the end result is identical.


src/schemaType/DocumentArrayType.ts
```diff
@@ -96,7 +105,7 @@ class DocumentArrayType extends BaseSchemaType {
 	}
 
 	/** Generate subdocument instances */
-	private *makeSubDocument(record: MvRecord): Generator<Document> {
+	private *makeSubDocument(record: MvRecord): Generator<Document<TSchema, TSchemaDefinition>> {
```
philfuster, line 108: You learn something new every time. This is my first time noticing the use of the `*` operator here and had to read up on `Generator`s. I could see this being very helpful for large `DEPOSITS` records.

philfuster (commented):
This looks really good. Lots to learn here about the things you can do with TypeScript types. Thank you for requesting my review.
I just had a few questions. Thank you.

src/Schema.ts
```diff
@@ -88,10 +87,66 @@ interface DictionaryTypeDetail {
...
+		? {
+				[K in keyof TSchemaDefinition]: InferSchemaType<TSchemaDefinition[K]>;
+			}
+		: never;
+
+/** Infer the shape of a `Model` instance based upon the Schema it was instantiated with */
+export type InferModelObject<TSchema extends Schema<SchemaDefinition>> = {
+	_id: string;
+	__v: string;
+} & InferDocumentObject<TSchema> extends infer O
```
kthompson23, line 143: Does it make sense to have a special Model type for `_raw` records?

Since there's no Schema how dos this inference work?

kthompson23 (commented):
This looks great but I think we need special handling for `_raw` type records. In those cases the Schema is null so requires `_raw` to be defined somewhere.

src/__tests__/Schema.test.ts
```diff
@@ -376,3 +378,767 @@ describe('transformPathsToDbPositions', () => {
...
+					{ isoCalendarDateProp: `${number}-${number}-${number}` }
+				> = true;
+				expect(test2).toBe(true);
+			});
+
+			test('should infer ISOTime type', () => {
+				const schema1 = new Schema({ isoTimeProp: { type: 'ISOTime', path: '1' } });
+				const test1: Assert<
+					InferDocumentObject<typeof schema1>,
+					{ isoTimeProp: `${number}:${number}:${number}.${number}` | null }
```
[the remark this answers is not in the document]
shawnmcknight: I've added an exported type for each of these types and used them where possible.

src/__tests__/Schema.test.ts
```diff
@@ -376,3 +378,767 @@ describe('transformPathsToDbPositions', () => {
 		expect(schema.transformPathsToDbPositions(['not.here'])).toEqual([]);
 	});
 });
+
+describe('utility types', () => {
```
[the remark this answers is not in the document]
shawnmcknight: I don't want to make a bunch of tests because these aren't real tests. They're just a container to keep the type assertions in.

I did add comments to each individual test section.

src/Schema.ts
shawnmcknight: I made some changes to improve the experience of schema-less models and documents in a794f2bbb09a806bae03d00e13fadf04d0d92766. It wasn't really broken previously -- it just ended up using a generic `SchemaDefinition` which couldn't be used to infer anything at all. However, it could be improved, particularly around the behavior of the `_raw` property to ensure it is typed appropriately based on whether a schema is supplied or not.

I further added tests in 0a3cb8d41cda648085a030c7e379ba1ee2a8d443 to confirm the output types for `Document` and `Model` methods and included both schema-less and schema variants in them.

philfuster (approved):
Your changes look good to me! Thank you.

kthompson23 (approved):
Looks good, thank you.

PR #701, opened 2024-06-19
Make validators synchronous to improve performance

shawnmcknight:
Currently, the document and schema type validators are asynchronous. When validating a document, this could result in an incredibly large number of asynchronous validations and therefore an equivalently large number of Promises. The volume of validations goes hand in hand with the total number of properties in a Document which, especially when factoring in DocumentArrays, might number in the many thousands or more.

However, none of the existing validators are asynchronous. The async functionality was created for a future use case that, to date, has not come.

I did a JS performance test and found that an array of 1000 items can be processed 217x faster with a synchronous function vs. an identical async function that doesn't do anything requiring async. Therefore, the supporting of async is reducing performance without any benefits.

This PR changes the `Document` and various schema type validations to no longer be async. If, in the future, async validation becomes necessary, we should consider a mechanism where the async validators will be kept independent of any synchronous validators.

### DIFF
--- a/src/Document.ts
+++ b/src/Document.ts
@@ -229,7 +229,7 @@ class Document<
 	}
 
 	/** Validate document for errors */
-	public async validate(): Promise<Map<string, string[]>> {
+	public validate(): Map<string, string[]> {
 		const documentErrors = new Map<string, string[]>();
 
 		if (this.#schema !== null) {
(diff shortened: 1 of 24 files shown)


shawnmcknight:
Jsperf test: https://jsperf.app/kixare

philfuster (approved):
Looks good to me! Thank you.

kthompson23 (approved):
Looks good, thank you. I think this will also help a lot with memory usage.

PR #719, opened 2024-07-09
Add type inference for query conditions

shawnmcknight:
This PR modifies the `Query` class so that the query conditions supplied to its constructor can have their types inferred from the `Schema` the query is operating against. The following sections will outline the changes made to support this.

#### `Schema` module and class
- Since queries come from a combination of schema properties and additional schema dictionaries, a new generic has been added to the schema class to infer the additional schema dictionaries. This allows access to those virtual dictionaries in the schema.

- A new `FlattenDocument` helper type was added to the module's exported types. It will construct an object type that is keyed by the string keypath (i.e. `foo.bar.baz`) of a document's nested properties with a value that corresponds to the scalar type at that path.

- A new `InferSchemaPaths` helper type was added to the module's exported types. It will return a union of the keypaths of its flattened document. `InferSchemaPaths` isn't technically necessary for this PR, but it could have utility in other use cases where the keypath is all that is necessary.

- New exported `DictionaryTypeDefinition...` types were created so that those could be used for constraint checking.

- The `TSchemaDefinition` generic has been given a default value of `SchemaDefinition`. This turned out to not cause any ill effects with inference and allowed significantly easier use of the `Schema` type elsewhere in the codebase since it wasn't required to supply the generic value.

#### `Query` module and class
- The `Query` module's `Filter` type has been adjusted to utilize the `FlattenDocument` type. This type will now produce conditions for each property in the flattened document which allows for type safety when constructing query conditions. The `Filter` will now enforce the correct keys as well as data types for those keys in the conditions, taking into account the flattened keypaths for the condition keys.

- Additionally, the `Filter` type makes use of a new local `InferDictionariesType` helper type which will infer the additional dictionaries supplied to the schema and include those as eligible filter properties.

- The `SortCriteria` type has been adjusted so that it uses the same queryable properties as defined by the `Filter` type by extracting its keys.

#### Other changes
An unrelated change has also been made to eliminate the use of the `GenericObject` type and replacing it with more appropriate types.

The `Assert` helper type proved to not work properly when attempting to test the `Filter` utility type. A new much more robust type equality checking was added to a new `Equals` type module.

Utility types to flatten objects were added to the `Utility` types module. This was used to produce the flattened document type.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,8 +28,8 @@
         "@types/lodash": "4.17.6",
         "@types/node": "20.14.10",
         "@types/nunjucks": "3.2.6",
-        "@typescript-eslint/eslint-plugin": "7.15.0",
-        "@typescript-eslint/parser": "7.15.0",
+        "@typescript-eslint/eslint-plugin": "7.16.0",
+        "@typescript-eslint/parser": "7.16.0",
         "babel-jest": "29.7.0",
         "babel-plugin-module-resolver": "5.0.2",
         "eslint": "8.57.0",
@@ -45,7 +45,7 @@
         "npm-run-all2": "6.2.2",
         "nunjucks": "3.2.4",
         "prettier": "3.3.2",
-        "rimraf": "5.0.8",
+        "rimraf": "5.0.9",
         "typescript": "5.5.3"
       },
       "engines": {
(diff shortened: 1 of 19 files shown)


shawnmcknight:
The inspiration for the `FromPaths` and `ToPaths` helpers came from https://dev.to/tylim88/typescript-type-level-object-flattening-9fn. It had to be tweaked a bit to avoid the infinite recursion and to handle arrays better, along with some separate small modifications to make it work a little better.

philfuster (approved):
This looks good to me! Thanks.

kthompson23 (commented):
@shawnmcknight This is a really great idea but does it work when querying on dictionaries?

ex: We query the PRODUCT file against `sku` but that is a virtual dictionary and not a property defined in the schema. It's defined in the `dictionaries` property.

shawnmcknight:
> @shawnmcknight This is a really great idea but does it work when querying on dictionaries?
> 
> ex: We query the PRODUCT file against `sku` but that is a virtual dictionary and not a property defined in the schema. It's defined in the `dictionaries` property.

Good point. I need to augment this to support that.

shawnmcknight:
> @shawnmcknight This is a really great idea but does it work when querying on dictionaries?
> 
> ex: We query the PRODUCT file against `sku` but that is a virtual dictionary and not a property defined in the schema. It's defined in the `dictionaries` property.

I've added support for inferring the virtual dictionaries as well.

philfuster (commented):
> Utility types to flatten objects were added to the Utility types module. This was used to product the flattened document type.

This is the last sentence of the PR description. `product` may be wrong. I'm thinking it might have meant to be `produce`.

philfuster (commented):
I wonder if the use of  `Remap` in certain scenarios, for example with the new `DictionaryTypeDefinitionString` types and the otehr dictionary related types, would be helpful for readability.

philfuster (approved):
This LGTM! This really is a master class in the power of typescript types. Great work. I thought it was pretty gnarly with the addition of FlattenDocument and what not, but then the Equals module is pretty crazy too. I think it would be good to go over this again in the next Back-end sync.

kthompson23 (approved):
Looks good, thank you.

PR #720, opened 2024-07-10
Eliminate unnecessary TSchemaDefinition generics

shawnmcknight:
While doing the work for #719, I learned that the `TSchemaDefinition` generics being used in several types/classes/etc. is actually not necessary. I think it was originally necessary in a lot of those scenarios but further refinements in the initial PR that utilized them made it ultimately unnecessary. When removing that generic from these places, the inference continues to work properly and the types are significantly easier to work with.

This PR eliminates all of those `TSchemaDefinition` generics that were in place except in the `Schema` class itself where the generic is used to obtain the definition supplied to the `Schema` constructor.

### DIFF
--- a/src/Connection.ts
+++ b/src/Connection.ts
@@ -350,10 +350,10 @@ class Connection {
 	}
 
 	/** Define a new model */
-	public model<
-		TSchema extends Schema<TSchemaDefinition> | null,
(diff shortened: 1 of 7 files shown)


kthompson23 (approved):
Looks good, thank you.

PR #724, opened 2024-07-15
Include cause in MvisError, TimeoutError, and UnknownError

shawnmcknight:
As of Node 16.9+, the `cause` property is available on `Error` objects. This allows for an improvement with scenarios where an error is caught and thrown as a different type of error so that the original error can be included on the thrown `Error`.

This PR modifies the `MvisError`, `TimeoutError`, and `UnknownError` subclasses so that a `cause` parameter is required. The provided `cause` will be set as the `cause` property on the `Error` instance which will allow consumers to obtain the originating error.

For `MvisError` and `TimeoutError`, the `cause` will always be an `AxiosError`. For `UnknownError`, the `cause` is of type `unknown`.

As part of this change, the `mvisRequest` and `mvisResponse` properties have been removed from the `MvisError` instance as this information is available on the `cause` property.

### DIFF
--- a/src/Connection.ts
+++ b/src/Connection.ts
@@ -482,26 +482,22 @@ class Connection {
 	private handleAxiosError(err: AxiosError): never {
 		if (err.code === 'ETIMEDOUT') {
 			this.logHandler.error(`Timeout error occurred in MVIS request: ${err.message}`);
-			throw new TimeoutError({ message: err.message });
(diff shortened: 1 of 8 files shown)


philfuster (approved):
This looks good to me! Thank you.

kthompson23 (approved):
Looks good, thank you.

PR #750, opened 2024-08-05
Mvom increment

philfuster:
### Summary

This PR implements the new `increment` operation.

#### TypeScript Changes

- Created Input and Output types for new `increment` operation
- Created async static method `increment` on dynamically generated Model class via `compileModel`.
- Created `transformPathToOrdinalPosition` to transform an increment operation input key path to its ordinal position.
- Updated `Connection` to handle new `RecordNotFoundError`.

#### UniBasic Changes

- Created unibasicTemplate - `increment.njk`
    - validates increment operation input
    - Has a read lock loop to handle locked records.
        -  sleeps according to the passed in retry delay.
        - returns `RecordLockedError` when retry max is exceeded
    - returns new `RecordNotFoundError` error when `else` statement executes on `readu`.
    -  writes record after performing incrementing.
    - returns the updated document
  
### Testing Notes

Ran tests using MVOM:

- incrementing a record that does not exist - received `RecordNotFoundError`
- incrementing one field - field was incremented, updated document returned
- incremented multiple fields - fields were incremented, updated document returned
- incremented nested field - nested field was incremented, updated document returned
- incremented value position - value position was incremented
- incremented multiple value positions - value positions were incremented
- incremented sub value positions - sub value positions were incremented
- incremented while record was locked - retries were performed - `RecordLockedError` returned.

### DIFF
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -180,7 +180,10 @@ export type FlattenDocument<TSchema extends Schema, TConstraint = SchemaTypeDefi
 		: never;
 
 /** Infer the string keyPaths of a schema */
-export type InferSchemaPaths<TSchema extends Schema> = keyof FlattenDocument<TSchema>;
+export type InferSchemaPaths<TSchema extends Schema> = Extract<
+	keyof FlattenDocument<TSchema>,
+	string
+>;
 // #endregion
 
 /** Schema constructor */
@@ -283,7 +286,7 @@ class Schema<
 	 * @throws {Error} Invalid path provided
 	 */
 	public transformPathToOrdinalPosition(
-		path: Extract<InferSchemaPaths<Schema<TSchemaDefinition, TDictionaries>>, string>,
+		path: InferSchemaPaths<Schema<TSchemaDefinition, TDictionaries>>,
 	): `${number}.${number}.${number}` {
 		const positionPaths = this.getPositionPaths();
 
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -278,6 +278,61 @@ class Schema<
 		return [...positions];
 	}
 
+	/** Transform the paths to ordinal positions. Returning a '.' delimited string. Ex. "1.2.3"  */
+	public transformPathsToOrdinalPositions(paths: string[]): string[] {
+		if (paths.length === 0) {
+			return [];
+		}
+
+		const positionPaths = this.getPositionPaths();
+		const positionKeys = Array.from(positionPaths.keys());
+
+		const positions = paths.reduce((acc, positionPath) => {
+			if (positionPaths.has(positionPath)) {
+				// find the key in position paths
+				// add position
+				const ordinalPath =
+					// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+					positionPaths.get(positionPath)!;
+				acc.add(this.buildOrdinalPathString(ordinalPath));
+			} else if (!positionPath.includes('.')) {
+				// if the property is a parent key, we will add positions for all children
+				// e.g we only pass property "name" to return all data for name.first, name.last, etc.
+				const matchedPositionPaths = positionKeys.filter(
+					(key) => key.split('.')[0] === positionPath,
+				);
+				matchedPositionPaths.forEach((key) => {
+					// add child position
+					// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+					const ordinalPath = positionPaths.get(key)!;
+					acc.add(this.buildOrdinalPathString(ordinalPath));
+				});
+			}
+			return acc;
+		}, new Set<string>());
+
+		return [...positions];
+	}
+
+	private buildOrdinalPathString(ordinalPath: number[]): string {
+		if (ordinalPath.length === 0) {
+			return '';
+		}
+
+		const [attributePosition, valuePosition, subvaluePosition] = ordinalPath;
+		if (ordinalPath.length === 1) {
+			return `${attributePosition + 1}`;
+		}
+		if (ordinalPath.length === 2) {
+			return `${attributePosition + 1}.${valuePosition + 1}`;
+		}
+		if (ordinalPath.length === 3) {
+			return `${attributePosition + 1}.${valuePosition + 1}.${subvaluePosition + 1}`;
+		}
+
+		return '';
+	}
+
 	/** Build the dictionary path map for additional dictionaries provided as schema options */
 	private buildDictionaryPaths(
 		dictionaries: DictionariesOption = {},
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -279,35 +279,23 @@ class Schema<
 	}
 
 	/**
-	 * Transform the paths to ordinal positions. Returning a '.' delimited string. Ex. "1.2.3"
-	 * @throws {Error} No paths provided
+	 * Transform the path to its ordinal position. Returning a '.' delimited string. Ex. "1.2.3"
 	 * @throws {Error} Invalid path provided
 	 */
-	public transformPathsToOrdinalPositions(
-		paths: Extract<InferSchemaPaths<Schema<TSchemaDefinition, TDictionaries>>, string>[],
-	): `${number}.${number}.${number}`[] {
-		if (paths.length === 0) {
-			throw new Error('No paths provided');
-		}
-
+	public transformPathToOrdinalPosition(
+		path: Extract<InferSchemaPaths<Schema<TSchemaDefinition, TDictionaries>>, string>,
+	): `${number}.${number}.${number}` {
 		const positionPaths = this.getPositionPaths();
 
-		const positions = paths.reduce((acc, positionPath) => {
-			if (positionPaths.has(positionPath)) {
-				// find the key in position paths
-				// add position
-				const ordinalPath =
-					// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-					positionPaths.get(positionPath)!;
-				const [attributePosition, valuePosition = 0, subvaluePosition = 0] = ordinalPath;
-				acc.add(`${attributePosition + 1}.${valuePosition + 1}.${subvaluePosition + 1}`);
-			} else {
-				throw new Error('Invalid path provided');
-			}
-			return acc;
-		}, new Set<`${number}.${number}.${number}`>());
+		if (!positionPaths.has(path)) {
+			throw new Error('Invalid path provided');
+		}
 
-		return [...positions];
+		const ordinalPath =
+			// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+			positionPaths.get(path)!;
+		const [attributePosition, valuePosition = 0, subvaluePosition = 0] = ordinalPath;
+		return `${attributePosition + 1}.${valuePosition + 1}.${subvaluePosition + 1}`;
 	}
 
 	/** Build the dictionary path map for additional dictionaries provided as schema options */
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -278,14 +278,19 @@ class Schema<
 		return [...positions];
 	}
 
-	/** Transform the paths to ordinal positions. Returning a '.' delimited string. Ex. "1.2.3"  */
-	public transformPathsToOrdinalPositions(paths: string[]): string[] {
+	/**
+	 * Transform the paths to ordinal positions. Returning a '.' delimited string. Ex. "1.2.3"
+	 * @throws {Error} No paths provided
+	 * @throws {Error} Invalid path provided
+	 */
+	public transformPathsToOrdinalPositions(
+		paths: Extract<InferSchemaPaths<Schema<TSchemaDefinition, TDictionaries>>, string>[],
+	): `${number}.${number}.${number}`[] {
 		if (paths.length === 0) {
-			return [];
+			throw new Error('No paths provided');
 		}
 
 		const positionPaths = this.getPositionPaths();
-		const positionKeys = Array.from(positionPaths.keys());
 
 		const positions = paths.reduce((acc, positionPath) => {
 			if (positionPaths.has(positionPath)) {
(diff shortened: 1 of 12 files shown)


src/Schema.ts
```diff
@@ -277,6 +277,61 @@ class Schema<
 		return [...positions];
 	}
 
+	/** Transform the paths to ordinal positions. Returning a '.' delimited string. Ex. "1.2.3"  */
+	public transformPathsToOrdinalPositions(paths: string[]): string[] {
```
shawnmcknight, line 281: This method feels weird for several reasons:
1. You are accepting a `string[]` and returning a `string[]` but the only use case calls for passing a single value in and getting a single value back.
2. The logic in the `else if` seems improper. For the projection use case, if a parent key was supplied then it needed to also produce paths for all children. In this use case, the expectation would be that a full keypath is supplied and not partial keypaths.
3. The `paths` isn't really a `string[]`. It's specifically the keys from the flattened document structure.
4. The return type is specifically `${number} | ${number}.${number} | ${number}.${number}.${number}`.
5. If the path isn't found then returning empty string or potentially an empty list seems improper. That would then set the `path` that gets passed to basic as an empty string or undefined value and then cause problems on the basic side.

src/Schema.ts
```diff
@@ -277,6 +277,61 @@ class Schema<
...
+					acc.add(this.buildOrdinalPathString(ordinalPath));
+				});
+			}
+			return acc;
+		}, new Set<string>());
+
+		return [...positions];
+	}
+
+	private buildOrdinalPathString(ordinalPath: number[]): string {
```
shawnmcknight, line 316: This feels like we're relying on a quirk of UniBasic where it will let you have empty strings for values and subvalues. It also means this code is more complex than it needs to be. It could simply be:

```
		const [attributePosition, valuePosition = 0, subvaluePosition = 0] = ordinalPath;
		return `${attributePosition + 1}.${valuePosition + 1}.${subvaluePosition + 1}`;
```

That code is then simple enough that it can probably just be consolidated into the `transformPathsToOrdinalPositions` method (especially since I think that can be slimmed down) and it won't need additional checks for length and such.

shawnmcknight (changes requested):
I've finished a first pass review and left several comments. Overall this looks good, but I had suggestions for improvement.

src/Schema.ts
philfuster: > 3. The paths isn't really a string[]. It's specifically the keys from the flattened document structure.
I'm trying to get this to work, but I'm struggling with how can I pass the Schema to `FlattenDocument` or `InferSchemaPath` when I'm within Schema. I've tried passing `this` and `typeof this` as the type parameter, but that wasn't working.

src/compileModel.ts
```diff
@@ -51,6 +56,25 @@ export interface ModelFindByIdOptions extends ModelDatabaseExecutionOptions {
...
 	projection?: string[];
 }
+export interface ModelIncrementOptions extends ModelDatabaseExecutionOptions {
+	/**
+	 * number retries to perform when record is locked
+	 * @defaultValue 5
+	 */
+	retry?: number;
+	/**
+	 * between retries in seconds when record is locked
```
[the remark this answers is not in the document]
philfuster: changed

src/compileModel.ts
```diff
@@ -51,6 +56,25 @@ export interface ModelFindByIdOptions extends ModelDatabaseExecutionOptions {
 	/** Array of projection properties */
 	projection?: string[];
 }
+export interface ModelIncrementOptions extends ModelDatabaseExecutionOptions {
+	/**
+	 * number retries to perform when record is locked
```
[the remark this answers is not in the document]
philfuster: done

src/compileModel.ts
```diff
@@ -288,6 +312,47 @@ const compileModel = <TSchema extends Schema | null>(
...
+		 * @throws {Error} if schema is not defined
+		 * @throws {Error} if no result is returned from increment operation
+		 */
+		public static async increment(
+			id: string,
+			operations: IncrementOperation<TSchema>[],
+			options: ModelIncrementOptions = {},
+		): Promise<ModelCompositeValue<TSchema>> {
+			const { maxReturnPayloadSize, requestId, userDefined, retry = 5, retryDelay = 1 } = options;
+			if (this.schema == null) {
```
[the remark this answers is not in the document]
philfuster: makes sense. removed the check

src/compileModel.ts
```diff
@@ -326,6 +391,33 @@ const compileModel = <TSchema extends Schema | null>(
...
+		static #formatIncrementOperations(
+			operations: IncrementOperation<TSchema>[],
+		): DbSubroutineInputIncrementOperation[] {
+			if (this.schema == null) {
+				// should never get here because increment also checks for null schema, but just in case.
+				throw new Error('Schema must be defined to perform increment operations');
+			}
+
+			return operations.map(({ path, value }) => {
+				if (this.schema == null) {
```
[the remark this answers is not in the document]
philfuster: good idea, thank you. done.

src/Schema.ts
shawnmcknight: This worked for me:
```
	/** Transform the paths to ordinal positions. Returning a '.' delimited string. Ex. "1.2.3"  */
	public transformPathsToOrdinalPositions(
		paths: Extract<
			keyof FlattenDocument<typeof this, Schema | SchemaDefinition | SchemaTypeDefinitionNumber>,
			string
		>[],
	): string[] {
	```
Although I would recommend making an alias for that and using it in both places it would be needed.

File: src/unibasicTemplates/subroutines/external/increment.njk [hunk 7]
```diff
@@ -0,0 +1,185 @@
...
+  * get subvalue position
+  subvaluePosition = field(path, '.', 3, 1)
+
+  * perform incrementing
+  record<attributePosition, valuePosition, subvaluePosition>+= value
+repeat
+
+* create null projection
+if udoCreate(UDO_NULL, projection) then
+   call error_handler(ERROR_UDO, output)
```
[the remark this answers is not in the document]
philfuster: done. Fixed in save as well.

src/unibasicTemplates/subroutines/external/increment.njk
```diff
@@ -0,0 +1,185 @@
...
+    call error_handler(ERROR_MALFORMED_INPUT, output)
+    go closeReleaseAndReturnFromSub
+  end
+  * value must be numeric
+  if type ne UDO_NUMBER then
+    call error_handler(ERROR_MALFORMED_INPUT, output)
+    go closeReleaseAndReturnFromSub
+  end
+
+  * parse out path.
```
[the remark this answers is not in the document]
philfuster: Good idea. Done.

src/Schema.ts
philfuster: I was getting infinite type instantiation as well

shawnmcknight:
```
	public transformPathsToOrdinalPositions(
		path: InferSchemaPaths<Schema<TSchemaDefinition>>,
	): string {
	```
	
	Since that method is on `Schema` it doesn't require the same restrictions to the paths that the increment method requires. So I think this will work.

Comment (philfuster) [reply src/Schema.ts:282]: That works but then the type of the paths is a union of  `string | number | symbol`. I'm casting the value to `String`, but I'm not 100% sure that's accurate. I'm not entirely sure why it's a union of the 3 types either.

File: src/Schema.ts (hunk 2 above)
Comment (philfuster) [reply src/Schema.ts:317]: Consolidated into `transformPathsToOrdinalPositions`

File: src/Schema.ts (hunk 1 above)
Comment (philfuster) [reply src/Schema.ts:282]: Interesting I would not have thought to just extract the string type from that. Why does that work? It makes sense what Extract does, but why is it valid/the right thing to do, to just ignore the other types?

I wrap `InferSchemaPaths` as such: `Extract<InferSchemaPaths<Schema<TSchemaDefinition>>, string>` and that appears to work within `transformPathsToOrdinalPositions`, but I end up with `never` when using `transformPathsToOrdinalPositions`.
image

Comment (philfuster) [reply src/Schema.ts:282]: I'm thinking it has to do with the `TDictionaries` as it defaults to `Record<never, never>`

Comment (philfuster) [reply src/Schema.ts:282]: but passing `TDictionaries` along does not help.

File: src/types/DbSubroutineOutput.ts [hunk 9]
```diff
@@ -31,6 +31,11 @@ export interface DbSubroutineOutputFindByIds {
 }
 export type DbSubroutineResponseFindByIds = DbSubroutineResponse<DbSubroutineOutputFindByIds>;
 
+export interface DbSubroutineOutputIncrement {
+	result: DbDocument | null;
```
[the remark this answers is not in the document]
philfuster: I don't think it is. I will remove `null` form the output type.

src/compileModel.ts
```diff
@@ -288,6 +312,47 @@ const compileModel = <TSchema extends Schema | null>(
...
+					retry,
+					retryDelay,
+				},
+				{
+					...(maxReturnPayloadSize != null && { maxReturnPayloadSize }),
+					...(requestId != null && { requestId }),
+					...(userDefined && { userDefined }),
+				},
+			);
+			if (data.result == null) {
```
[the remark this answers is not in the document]
philfuster: yea I don't think `null` is a valid response. I have removed this check.

src/unibasicTemplates/subroutines/external/increment.njk
```diff
@@ -0,0 +1,185 @@
...
+repeat
+
+* create null projection
+if udoCreate(UDO_NULL, projection) then
+   call error_handler(ERROR_UDO, output)
+   go closeAndReturnFromSub
+end
+
+* write out record
+write record on f.file, recordId on error
```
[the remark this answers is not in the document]
philfuster: Created this jira task - https://storis.atlassian.net/browse/CXM-8262

philfuster:
> I've finished a first pass review and left several comments. Overall this looks good, but I had suggestions for improvement.

@shawnmcknight Thank you for your review. I have addressed your feedback in the latest series of commits. This is ready for another look.

src/unibasicTemplates/subroutines/external/increment.njk
```diff
@@ -0,0 +1,189 @@
...
+  end
+  * value must be numeric
+  if type ne UDO_NUMBER then
+    call error_handler(ERROR_MALFORMED_INPUT, output)
+    go closeReleaseAndReturnFromSub
+  end
+
+  * parse out path.
+  pathMatchTest = path match "1n'.'1n'.'1n"
+  if pathMatchTest eq 0 then
```
shawnmcknight, line 122: This is returning a 1 or a 0 which is a boolean true/false in basic. It should be treated as boolean for clarity.

You can consolidate this into a single line expression:
`if not(path matches "1n'.'1n'.'1n") then`

src/compileModel.ts
```diff
@@ -51,6 +56,25 @@ export interface ModelFindByIdOptions extends ModelDatabaseExecutionOptions {
...
+	/**
+	 * Delay between retries in seconds when record is locked
+	 * @defaultValue 1
+	 */
+	retryDelay?: number;
+}
+
+export type IncrementOperation<TSchema extends Schema | null> = TSchema extends Schema
+	? {
+			path: keyof FlattenDocument<TSchema, Schema | SchemaDefinition | SchemaTypeDefinitionNumber>;
```
shawnmcknight, line 74: This should also have the `Extract` on it since the called method expects it.

src/__tests__/Schema.test.ts
```diff
@@ -388,6 +388,67 @@ describe('transformPathsToDbPositions', () => {
...
+			{
+				docStringProp: { type: 'string', path: '10' },
+				docNumberProp: { type: 'number', path: '11' },
+			},
+		],
+		documentArraySchemaProp: [new Schema(documentArrayDefinition)],
+	} satisfies SchemaDefinition;
+	const schema = new Schema(definition);
+
+	test('should throw error if path list is empty', () => {
```
shawnmcknight, line 422: This test description doesn't seem to align with the test.

shawnmcknight (changes requested):
Looks good. Just a few minor comments left.

src/Schema.ts
```diff
@@ -278,6 +278,26 @@ class Schema<
...
+	/**
+	 * Transform the path to its ordinal position. Returning a '.' delimited string. Ex. "1.2.3"
+	 * @throws {Error} Invalid path provided
+	 */
+	public transformPathToOrdinalPosition(
+		path: Extract<InferSchemaPaths<Schema<TSchemaDefinition, TDictionaries>>, string>,
+	): `${number}.${number}.${number}` {
+		const positionPaths = this.getPositionPaths();
+
+		if (!positionPaths.has(path)) {
```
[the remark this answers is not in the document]
philfuster: Agreed. done.

src/Schema.ts
```diff
@@ -278,6 +278,26 @@ class Schema<
 		return [...positions];
 	}
 
+	/**
+	 * Transform the path to its ordinal position. Returning a '.' delimited string. Ex. "1.2.3"
+	 * @throws {Error} Invalid path provided
+	 */
+	public transformPathToOrdinalPosition(
+		path: Extract<InferSchemaPaths<Schema<TSchemaDefinition, TDictionaries>>, string>,
```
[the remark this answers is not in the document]
philfuster: Yea that makes a lot more sense. Done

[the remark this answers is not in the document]
philfuster: Do you know how `number` and `symbol` end up in that resulting type?

src/__tests__/Schema.test.ts
philfuster: done. Thank you.

src/compileModel.ts
philfuster: Done. added Extract here.

src/unibasicTemplates/subroutines/external/increment.njk
philfuster: Done thank you.

shawnmcknight (approved):
LGTM!

PR #753, opened 2024-08-06
Fix unreported error condition in save subroutine

shawnmcknight:
In the `save` basic subroutine there is an error check if the `digest` function fails. However, this code is not properly handling the error and is merely exiting without finishing the execution of the `save` subroutine or recording the error.

This PR updates the `save` subroutine so that the error will be properly recorded.


kthompson23 (approved):
Looks good, thank you.

PR #754, opened 2024-08-06
Change _raw from conditionally undefined to conditionally never

shawnmcknight:
#697 modified the `_raw` property of a document so that it was only conditionally undefined if the document did not have a `Schema` associated with it. However, this change had the side-effect of making the `_raw` property an object key in 100% of document instances whereas it was previously only conditionally added as a property. There is a subtle nuance to a property not being present and being present but undefined, and this exposed that nuance.

This PR changes things slightly such that the `_raw` property will only be assigned a value (and thus only become an object property) if there is no schema. If a schema is present then the type of `_raw` will be `never` and it will not be assigned.


kthompson23 (approved):
Looks good, thank you.

PR #756, opened 2024-08-07
Internal Write Subroutine to be Used in save and increment

philfuster:
### Summary

This PR creates an internal `write_record` subroutine to be used by both save and increment.
This prevents duplicated logic between the save and increment operations.

### Reviewers

@shawnmcknight


shawnmcknight (approved):
LGTM!

PR #757, opened 2024-08-07
Increment to return original and updated document

philfuster:
### Summary

The increment operation has been modified to return both the Original and Updated documents.
This prevents consumers from having to fetch the original document then perform the increment in order to have a copy of the original and updated document
This also prevents any issues related to the original document becoming out of sync with the updated document during the time between fetching the original document and performing the increment operation.

### Reviewers

@shawnmcknight


shawnmcknight (approved):
LGTM!

PR #760, opened 2024-08-09
Object Arrays Documentation

treisig:
There should be a single object with multiple fields inside the array to create associative properties, instead of multiple objects with a single field.


shawnmcknight (approved):
LGTM!

PR #789, opened 2024-09-19
Allow for required empty string enumerations

shawnmcknight:
Currently, you cannot have a string schema property which is both required and has an enumeration list containing empty string. Empty strings are considered as failing required validation so even though it is a permissible value as defined by the enumeration it will still trigger a required validation failure. However, on output transformation, if an empty string is in the enumeration list, the output from mvom will be empty string instead of `null`.

This PR modifies the required validation for strings so that if a property is required, it will not be considered an error if the value is present in the enumeration list, regardless of whether or not it is an empty string. This will allow empty strings to be defined for non-nullable fields and align the behavior more closely with the output transformation behavior for enumerations that contain empty strings.

### DIFF
--- a/src/schemaType/StringType.ts
+++ b/src/schemaType/StringType.ts
@@ -52,11 +52,12 @@ class StringType extends BaseScalarType {
 	}
 
 	/** String required validator */
-	protected override validateRequired = (value: unknown): boolean =>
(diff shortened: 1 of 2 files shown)


philfuster (approved):
LGTM!

PR #790, opened 2024-09-19
Add _id and __v to types for schemaless models

shawnmcknight:
The output of Model methods for models that are constructed without a schema should have `_id` and `__v` properties as required like models constructed with a schema. In the current state, the `_id` and `__v` properties of models without a schema is optional.

This PR augments the `ModelCompositeValue` type so that it makes the `_id` and `__v` properties as required.

### DIFF
--- a/src/compileModel.ts
+++ b/src/compileModel.ts
@@ -11,6 +11,7 @@ import type {
 	DbServerDelimiters,
 	DbSubroutineInputIncrementOperation,
 	DbSubroutineUserDefinedOptions,
+	Remap,
 } from './types';
 import { ensureArray } from './utils';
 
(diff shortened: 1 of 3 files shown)


src/compileModel.ts
```diff
@@ -28,13 +29,19 @@ export type ModelConstructor<TSchema extends Schema | null> = ReturnType<
 
 export type Model<TSchema extends Schema | null> = InstanceType<ModelConstructor<TSchema>>;
 
+/** Used as an intersection type to make _id & __v properties required */
+export interface RequiredModelMeta {
+	_id: string;
```
philfuster, line 34: I think `__v` could be `null` if it's a new record right? Would that mean this should be unioned with `null`? What about `_id`? Could it be `null`? Both class member variables are unioned with `null` on the dynamically created Model returned from `compileModel`.

philfuster (changes requested):
Looks good. I just had a question about whether `__v` and `_id` should be unioned with `null`. I feel like I might be missing something here so I figured I would ask.

src/compileModel.ts
philfuster: I see, thank you clearing that up for me! The tests help as well.

PR #794, opened 2024-09-23
Switch the type definition of array schema properties from arrays to readonly tuples

shawnmcknight:
This PR modifies the type definitions of "array" schema properties. Previously, these were defined as arrays. However, that is inaccurate as the schema type definitions do not allow specification of arrays of indefinite length -- they only allow arrays of length 1 (i.e. a tuple). The first change in this PR is to adjust these types to more accurately indicate that a 1-length tuple should be used when constructing array definitions.

A second change was also made to these definitions to designate those tuples as `readonly`. This is to offer better support for the use of `as const` when defining objects which are schema definitions. `as const` must be used for schema definitions that are created outside of the schema constructor parameters in order to get proper type inference for the schema definition. Without it, the generic for the schema is either an error or widens to the broad `SchemaDefinition` which cannot provide any actual type information from the schema. However, without allowing `readonly` for these arrays, the `as const` assertion itself would trigger a type error because it marks the tuples as readonly which is in violation of the defined type. By making the tuples `readonly`, a consumer can specify both readonly and writeable arrays in the schema definition, allowing for the proper use of `as const`.

### DIFF
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -48,10 +48,10 @@ export type SchemaTypeDefinition =
 	| SchemaTypeDefinitionArray;
 
 type SchemaTypeDefinitionArray =
-	| Schema[]
-	| SchemaTypeDefinitionScalar[]
-	| SchemaTypeDefinitionScalar[][]
-	| SchemaDefinition[];
+	| readonly [Schema]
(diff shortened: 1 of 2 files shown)


philfuster (approved):
Looks good! Thank you.

PR #805, opened 2024-10-16
Change output of Connection ISO date/time methods to use templated string literals

shawnmcknight:
The type inference for schema properties defined with the "ISO" types, `ISOCalendarDate`, `ISOTime`, and `ISOCalendarDateTime` all emit templated string literal types rather than simply strings. However, the `Connection` methods which return the current date, time, and date-time from the database are inconsistently returning a type of "string" from the database. Therefore, if you attempt to get a date, time, or date-time using the `Connection` methods and then supply that value to a schema property defined by one of those types it will cause a type error.

This PR adjusts the `Connection` methods `getDbDate`, `getDbDateTime`, and `getDbTime` to return `ISOCalendarDate`, `ISOCalendarDateTime`, and `ISOTime` types respectively. This makes those methods consistent with the types inferred from schemas.

Since both `Connection` and `Schema` now emit these types, they have been relocated to `types/Miscellaneous`.

### DIFF
--- a/src/Connection.ts
+++ b/src/Connection.ts
@@ -44,6 +44,9 @@ import type {
 	DbSubroutineResponseTypes,
 	DbSubroutineResponseTypesMap,
 	DbSubroutineSetupOptions,
+	ISOCalendarDate,
+	ISOCalendarDateTime,
+	ISOTime,
 } from './types';
 
 // #region Types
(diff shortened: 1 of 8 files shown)


philfuster (approved):
Looks good to me! Thank you.

PR #820, opened 2024-10-24
Improve inferred types for _raw property

shawnmcknight:
This PR makes improvements regarding the `_raw` property of `Document` instances (which also impacts `Model` since it extends `Document`).

The following changes are included:
- `DocumentData` will now include the `_raw` property if the schema is null
- `DocumentCompositeValue` will now include the `_raw` property if the schema is null
- `InferDocumentObject` now accepts a `null` schema in its generic. If the schema is `null` then the `_raw` property will be included on the inferred output
- `InferModelObject` now accepts a `null` schema in its generic.  If the schema is `null` then the `_raw` property will be included on the inferred output

### DIFF
--- a/src/Document.ts
+++ b/src/Document.ts
@@ -7,9 +7,9 @@ import type { DbServerDelimiters, DeepOptionalNullable, MvRecord } from './types
 
 // #region Types
 /** Type of data property for constructing a document dependent upon the schema */
-export type DocumentData<TSchema extends Schema | null> = TSchema extends Schema
-	? DeepOptionalNullable<InferDocumentObject<TSchema>>
-	: never;
+export type DocumentData<TSchema extends Schema | null> = DeepOptionalNullable<
+	InferDocumentObject<TSchema>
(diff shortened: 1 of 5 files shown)


shawnmcknight:
I need to think a bit more on the `_originalRecordString` side of this. I came to the realization that it is not an enumerable property. There is no way of conveying that in a TS type: https://github.com/microsoft/TypeScript/issues/9726.

That means consumers will think it's perfectly reasonable to iterate the object properties, or use the spread operator, and they would get the `_originalRecordString` when they would not. Making it enumerable could be a major breaking change.

The `Document` instance already has an index signature, although all properties are defined as `unknown`. This might simply need to be something that gets cast to in userland.

philfuster (changes requested):
This looks good to me! it's unfortunate it is not possible to convey to a user when a property is not enumerable, but casting in userland shouldn't be all that bad.

philfuster (approved):

PR #854, opened 2024-12-12
Allow equality, inequality, in, and not in query operators to accept null values

shawnmcknight:
This PR updates the query string formatting process to accept `null` as the conditional value for equality, inequality, in, and not in operators. For the other query operators, `null` does not make sense and could potentially be a security risk.

When parsing the query object, `null` will be converted to empty string such that query conditions will be compared against the empty string.

Schema fields indicated as required will continue to alert as type errors when using `null` comparisons as they should, in theory, return no results. Any additional dictionaries added to the schema will allow for `null` comparisons as there is no information available to indicate whether those dictionaries are nullable. These are only typechecking behaviors and will not affect the runtime behavior (i.e. `null` can be supplied at runtime for any of the permitted operators).

One additional minor change was made to a few parameters with array values. The type of these were changed to `readonly` arrays as they do not mutate the input and should allow consumers to supply both readonly and mutable arrays, offering more flexibility.

### DIFF
--- a/src/__tests__/Query.test.ts
+++ b/src/__tests__/Query.test.ts
@@ -1177,7 +1177,7 @@ describe('exec', () => {
 			});
 		});
 
-		test('should join multiple implicit conditions with and', async () => {
+		test('should join multiple implicit conditions with "and"', async () => {
 			const propertyName1 = 'property-name1';
 			const propertyValue1 = 'property-value1';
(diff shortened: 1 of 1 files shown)


philfuster (approved):
This looks good to me! Nice work. How'd this come about? Or was this always on your roadmap?

kthompson23 (approved):
Looks good, thank you.

PR #856, opened 2024-12-17
Fix type issue with filter/sort and arrays

shawnmcknight:
This PR fixes a type issue with the `Filter` and `SortCriteria` types when the schema includes scalar arrays without a dictionary specified in the definition. Any schema property without a dictionary should be excluded from the output of these utility types. However, due to a flaw in the constraint application in `InferDocumentObject`, the scalar definition of arrays is not being processed by the constraint check. This is because the constraint check in `InferDocumentObject` is on the entire property and is therefore bypassed for arrays. This caused the `Filter` and `SortCriteria`, which constrain to only those properties with a `dictionary` property, to errantly include scalar array properties without a dictionary defined.

The solution to this is to ensure that the scalar definition of a scalar array passes the supplied constraint. A new utility type `InferArraySchemaType` was added to assist in this which was then used in the existing `InferSchemaType` utility type.

New tests have been added to the utility type checks for `Filter` and `SortCriteria` to confirm this. In addition, tests were added for the `IncrementOperation` utility type to confirm that works as expected.

Finally, two unrelated changes were made. First, any variable declared as `SchemaDefinition` was changed to `satisfies SchemaDefinition` to ensure proper type inference of those objects. While testing, some new type errors appeared related to the use of `_id` in the `Document` class. The logic handling this property is inappropriately in this class as `_id` isn't a `Document` property but rather a `Model` property. I've added `TODO` notes to move this logic to the appropriate class at a later date.

### DIFF
--- a/src/Document.ts
+++ b/src/Document.ts
@@ -147,7 +147,10 @@ class Document<TSchema extends Schema | null> {
 				);
 	}
 
-	/** Build a list of foreign key definitions to be used by the database for foreign key validation */
+	/**
+	 * Build a list of foreign key definitions to be used by the database for foreign key validation
+	 * TODO: Foreign keys are a Model concept and not a Document concept. Relocate this logic.
+	 */
 	public buildForeignKeyDefinitions(): BuildForeignKeyDefinitionsResult[] {
 		if (this.#schema === null) {
 			return [];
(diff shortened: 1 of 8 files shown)


philfuster (approved):
This looks good to me! Thank you.

PR #859, opened 2024-12-24
Relocate ID checks from Document to Model

shawnmcknight:
A `Document` instance does not have an `_id` property. This property is specific to a `Model` instance. However, there exists validations on the `Document` instance to ensure that the `_id` has the correct pattern match as well as the construction of the foreign key definitions for the `_id` property.

This PR relocates the logic for validating an `_id` pattern and building the `_id` foreign key definitions from the `Document` class to the `Model` subclass.

Additionally, while doing this work, I noticed that embedded schemas are not properly building foreign key definitions as no logic was implemented to do so. The embedded schema type has been updated to build foreign key definitions.

### DIFF
--- a/src/__tests__/Document.test.ts
+++ b/src/__tests__/Document.test.ts
@@ -806,23 +806,6 @@ describe('buildForeignKeyDefinitions', () => {
 		expect(document.buildForeignKeyDefinitions()).toEqual(expected);
 	});
 
-	test('should build foreign key definitions for ids', () => {
-		const definition = {
-			prop1: { type: 'string', path: '1' },
-		} satisfies SchemaDefinition;
-		const schema = new Schema(definition, {
-			idForeignKey: { entityName: 'entityName', file: 'FILE' },
-		});
-
-		// @ts-expect-error: _id is not a valid Document property. See note regarding relocation in Document.ts.
-		const document = new DocumentSubclass(schema, { data: { _id: 'id', prop1: 'foo' } });
-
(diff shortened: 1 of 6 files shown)


src/compileModel.ts
```diff
@@ -490,6 +485,53 @@ const compileModel = <TSchema extends Schema | null>(
 				)
 				.join(am);
 		}
+
+		/** Validate the model instance */
+		#validate(): asserts this is { _id: string } {
```
philfuster, line 490: I had not seen this before, this is cool.

src/__tests__/Document.test.ts
```diff
@@ -701,6 +701,60 @@ describe('buildForeignKeyDefinitions', () => {
 		expect(document.buildForeignKeyDefinitions()).toEqual(expected);
 	});
 
+	test('should create foreign key definitions for nested object', () => {
```
philfuster, line 704: Did not building foreign key definitions for nested objects/embedded types impact anything in the services? I took a brief look and I don't think it did. I would imagine if it did we'd have gotten some bad data already.

philfuster (approved):
This looks really good. I only had some questions.

src/__tests__/Document.test.ts
shawnmcknight: There wasn't a problem with the FK and nested objects. There was a problem with embedded schemas, which I don't think have ever been used.

kthompson23 (approved):
Looks good, thank you.

PR #860, opened 2024-12-24
Improve database error types and handling

shawnmcknight:
This PR makes the following changes:

1. The `dbErrors` mapping is now a const object which utilizes the `satisfies` operator to ensure proper structure. This allows the object to be treated as its literal values rather than simply an arbitrary record.
2. New types `DbErrors` and `DbErrorCodes` have been created to create a type of the `dbErrors` object and a union of the possible error codes.
3. The error types from the db subroutine outputs have been tightened so that a foreign key validation error has the specific code value from the `dbErrors` mapping and other errors are all other possible codes.
4. Bullet 3 allowed for the removal of a type assertion in the `Connection` error handler as the code is now able to properly narrow the type for a foreign key error response.
5. Casting error codes to number has been removed as the output from the database subroutines is an error and not a string.
6. Unnecessary exports from the `DbSubroutineOutput` types has been eliminated.

### DIFF
--- a/src/Connection.ts
+++ b/src/Connection.ts
@@ -38,7 +38,6 @@ import type {
 	DbSubroutineInputIncrement,
 	DbSubroutineInputOptionsMap,
 	DbSubroutineInputSave,
-	DbSubroutineOutputErrorForeignKey,
 	DbSubroutinePayload,
 	DbSubroutineResponseError,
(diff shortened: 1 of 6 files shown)


kthompson23 (approved):
Looks good, thank you.

PR #904, opened 2025-03-31
Add a checkForRecordLockById method

stfstoris:
### Summary
This PR implements a new `checkForRecordLockById` method to provide a way to easily check if there is another process locking a record before attempting a mutation on record.

#### TypeScript Changes
- Create Input and Output types for new `checkForRecordLockById` method
- Create new static method `checkForRecordLockById` on dynamically created Model class in `compileModel`

#### UniBasic Changes
- Created unibasicTemplate `checkForRecordLockById.njk`
   - Opens given file by filename
   - Runs UniBasic function `RECORDLOCKED` and returns the numeric result. Method `checkForRecordLockedById` will then return a `boolean` value where true represents the record being locked (any value other than 0).

### Testing Notes
Ran the following tests:
- Checked a record in a file that was not locked - correctly returns `false`
- Checked a record that was locked by me - correctly returns `true`
- Checked a record that was locked by another process - correctly return `true`
- Checked a record that does not exist - correctly returns `false`

### DIFF
--- a/src/unibasicTemplates/subroutines/external/checkForRecordLockById.njk
+++ b/src/unibasicTemplates/subroutines/external/checkForRecordLockById.njk
@@ -34,7 +34,5 @@ if udoSetProperty(output, 'result', result) then
   go returnFromSub
 end
 
-go returnFromSub
-
 returnFromSub:
 return; * returning to caller
--- a/src/unibasicTemplates/subroutines/external/checkForRecordLockById.njk
+++ b/src/unibasicTemplates/subroutines/external/checkForRecordLockById.njk
@@ -15,34 +15,14 @@ if udoGetProperty(options, 'id', recordId, type) then
   go returnFromSub
 end
 
-* open the file
-open filename to f.file on error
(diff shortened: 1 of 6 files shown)


src/unibasicTemplates/subroutines/external/checkForRecordLockById.njk
```diff
@@ -0,0 +1,40 @@
...
+
+close f.file on error null
+
+* set the result property
+if udoSetProperty(output, 'result', result) then
+  call error_handler(ERROR_UDO, output)
+  go returnFromSub
+end
+
+go returnFromSub
```
shawnmcknight, line 37: This isn't necessary since the next line is the `returnFromSub` label.

shawnmcknight (changes requested):
This looks good to me. Just one small note about the extraneous go statement.

shawnmcknight (approved):
LGTM!

mattisax (approved):
Thanks @stfstoris!

PR #907, opened 2025-04-11
Update how dates are queried

stfstoris:
### Summary
This PR updates how properties typed as `ISOCalendarDate` are queried on in an effort to better handle dates that could be in different formats.

#### TypeScript Changes
Updates `ISOCalendarDateTransformer`'s `transformToQuery` method to now pass in a date query in the `ISOCalendarDate` format (yyyy-MM-dd) instead of the mv date (days since mv epoch)

#### UniBasic Changes
Updates `find.njk` subroutine to:
1. Capture the system `DATE.FORMAT` at beginning of subroutine 
2. Change `DATE.FORMAT` to `2` (international format) before executing query
3. Changes `DATE.FORMAT` back to whatever the system is set to after query executes

### Testing Notes
Tested that a query with a date parameter is now being queried as `select FOO with BAR >= "2025-04-01"` with date in `yyyy-MM-dd` format.

### DIFF
--- a/src/dataTransformers/ISOCalendarDateDataTransformer.ts
+++ b/src/dataTransformers/ISOCalendarDateDataTransformer.ts
@@ -47,9 +47,20 @@ class ISOCalendarDateDataTransformer implements DataTransformer {
 		return String(differenceInDays(this.parseISOCalendarDate(value), mvEpoch));
 	}
 
-	/** Transform query constants to internal u2 date format */
+	/** Transform query constants to ISOCalendarDate format */
 	public transformToQuery(value: unknown): string {
-		return value === '' || value == null ? '' : this.transformToDb(value);
+		if (value === '' || value == null) {
+			return '';
+		}
+
+		if (typeof value !== 'string') {
+			throw new TransformDataError({
+				transformClass: this.constructor.name,
+				transformValue: value,
+			});
+		}
+
+		return format(this.parseISOCalendarDate(value), ISOCalendarDateFormat);
 	}
 
 	/** Parse ISOCalendarDate string into date */
--- a/src/dataTransformers/ISOCalendarDateDataTransformer.ts
+++ b/src/dataTransformers/ISOCalendarDateDataTransformer.ts
@@ -60,7 +60,14 @@ class ISOCalendarDateDataTransformer implements DataTransformer {
 			});
 		}
 
-		return format(this.parseISOCalendarDate(value), ISOCalendarDateFormat);
+		try {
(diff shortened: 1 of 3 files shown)


src/dataTransformers/ISOCalendarDateDataTransformer.ts
```diff
@@ -47,9 +47,20 @@ class ISOCalendarDateDataTransformer implements DataTransformer {
...
+		}
+
+		if (typeof value !== 'string') {
+			throw new TransformDataError({
+				transformClass: this.constructor.name,
+				transformValue: value,
+			});
+		}
+
+		return format(this.parseISOCalendarDate(value), ISOCalendarDateFormat);
```
shawnmcknight, line 63: What happens here if the supplied `value` is not is `ISOCalendarDateFormat`? We probably want to ensure that someone can't supply something that would make the query perform unexpectedly as it might be somewhat similar to a SQL injection attack. If `parse` or `format` throws when an invalid input is encountered then it might be good to wrap this line in a `try/catch` block and re-throw the `TransformDataError`.

It would be worth having some tests of invalid inputs.

src/unibasicTemplates/subroutines/external/find.njk
```diff
@@ -31,13 +34,20 @@ if udoGetProperty(options, 'skip', skip, type) then
...
 * execute the query
 udtexecute queryCommand returning errorCodes
 if @system.return.code lt 0 then
   call error_handler(ERROR_QUERY, output)
+  execute 'date.format ':systemDateFormat
   go returnFromSub
 end
 
+* set date.format back
+execute 'date.format ':systemDateFormat
```
shawnmcknight, line 49: You could like set this back immediately after executing the query (and before the `@system.return.code` check). That way you wouldn't need to set it both when there was an error detected and without. That would also leave the least amount of lines of code between the initial setting and the reverting back.

shawnmcknight (changes requested):
Looks good. I had a question about how invalid inputs are handled (and some potential thoughts for improvement) and a suggestion for where to relocate the reverting of the date.format to.

src/dataTransformers/ISOCalendarDateDataTransformer.ts
stfstoris: Good call, I wrapped in `try/catch` and added a test to test some invalid values.

shawnmcknight:
Looks good! I will merge on Monday and cut a patch release.

shawnmcknight (approved):

PR #910, opened 2025-04-25
Force cast of error code to numeric

shawnmcknight:
The `error_handler` UniBasic subroutine adds a numeric error code to the output JSON. However, because of inconsistent behavior from `UDOSetProperty`, that numeric value will sometimes be added to the JSON as a numeric string rather than a number. This PR updates the logic for setting this property to force a cast to a number.


mattisax (approved):
Looks good!

PR #938, opened 2025-06-27
Add x-request-id headers to subroutine calls and request/trace headers to all requests

kthompson23:
This PR adds the `x-request-id` header to all subroutine calls. Many reverse proxies, including nginx, will generate a new requestId if this header is not provided. This change will allow users to link their mvom logs and reverse proxy logs.

I also added the `x-request-id` and `X-MVIS-Trace-id` headers if provided to all requests to mvis-admin. This allows mvis-admin logs and mvis-admin traces to also be linked as the request Id will appear in the the mvis-admin logs as the `OTEL-TRACE-ID`

### DIFF
--- a/src/Connection.ts
+++ b/src/Connection.ts
@@ -257,7 +257,7 @@ class Connection {
 
 		if (validateDeployment) {
 			this.logHandler.info('Validating deployment');
-			await this.validateDeployment();
+			await this.validateDeployment({ requestId });
 		} else {
 			this.logHandler.info('Skipping deployment validation');
 		}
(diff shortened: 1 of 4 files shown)


shawnmcknight:
> I also added the `x-request-id` and `X-MVIS-Trace-id` headers if provided to all requests to mvis-admin. This allows mvis-admin logs and mvis-admin traces to also be linked.

Can you clarify the usage of `X-MVIS-Trace-id` here? Will mvis admin be logging that? I'm not sure how this would be used.

shawnmcknight (changes requested):
This looks good, but I became confused about the `authenticate` method returning headers unrelated to authentication which were then used later in the process flow. I think the code is a bit hard to reason about now.

kthompson23:
Closing since this change is no longer needed.

PR #961, opened 2025-11-28
Fix 960 boolean field inference

### LINKED ISSUE #960: Fix boolean field type inference to match runtime behavior
The `InferModelObject` type has an issue where a `boolean` field that is not required is inferred as a union of `boolean` and `null`. This contradicts the runtime behvaior where mvom converts `null` to `false`.

Example: 

```ts
const fooSchema = new Schema({
  bar: { type: 'boolean', path: '268', dictionary: 'BAR' },
});

type fooData = InferModelObject<typeof fooSchema>;
// The type of fooData.bar is inferred as boolean | null;
```

This can lead to runtime issues if logic is taking a specific action based on the value being `null`.

mattisax:
Fixes #960

### Summary

This PR fixes TypeScript type inference for boolean schema fields to align with runtime behavior. Previously, boolean fields were typed as `boolean | null`, but the `BooleanDataTransformer` always converts `null` values to `false` at runtime, meaning boolean fields never actually return `null`. This change updates the type inference to correctly reflect that boolean fields always return `boolean`, eliminating a type-safety gap where TypeScript allowed `null` handling that would never execute.

### Technical Details

**Changed**: `InferSchemaType` type helper in `src/Schema.ts`
- Removed `InferRequiredType` wrapper from boolean type inference
- Boolean fields now directly infer as `boolean` regardless of the `required` property
- Other scalar types (string, number, date/time) continue to respect `required` and can still be `null`

### Breaking Changes

This is technically a breaking change for TypeScript consumers, though it corrects a type-safety issue:

- Code that previously handled `null` values for boolean fields will now show TypeScript errors
- However, such null checks were unnecessary since `null` is transformed to `false` at runtime
- Users needing true tri-state logic (true/false/null as distinct states) should use `string` type with `enum` constraint, as documented

### DIFF
--- a/src/__tests__/Query.test.ts
+++ b/src/__tests__/Query.test.ts
@@ -2403,7 +2403,7 @@ describe('utility types', () => {
 				{
 					$and?: readonly Filter<typeof schema>[];
 					$or?: readonly Filter<typeof schema>[];
-					booleanOptional?: Condition<boolean | null>;
+					booleanOptional?: Condition<boolean>;
 					booleanRequired?: Condition<boolean>;
 					stringOptional?: Condition<string | null>;
 					stringRequired?: Condition<string>;
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -145,7 +145,7 @@ type InferArraySchemaType<
 /** Infer the output type of a schema type definition */
 type InferSchemaType<TSchemaTypeDefinition, TConstraint> =
 	TSchemaTypeDefinition extends SchemaTypeDefinitionBoolean
-		? InferRequiredType<TSchemaTypeDefinition, boolean>
+		? boolean
 		: TSchemaTypeDefinition extends SchemaTypeDefinitionString
 			? InferRequiredType<TSchemaTypeDefinition, InferStringType<TSchemaTypeDefinition>>
 			: TSchemaTypeDefinition extends SchemaTypeDefinitionNumber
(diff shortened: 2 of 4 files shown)


src/__tests__/Query.test.ts
```diff
@@ -2403,7 +2403,7 @@ describe('utility types', () => {
 				{
 					$and?: readonly Filter<typeof schema>[];
 					$or?: readonly Filter<typeof schema>[];
-					booleanOptional?: Condition<boolean | null>;
+					booleanOptional?: Condition<boolean>;
```
shawnmcknight, line 2406: I wonder if there are existing scenarios where a query against an optional boolean is testing for both false and null to deal with the fact that basic will treat them equivalently.

shawnmcknight (approved):
Looks good. My only concern will be when it comes time to enable the use of this change. I wonder if any type errors will be encountered due to queries supplying `null` on "optional" boolean fields.

src/__tests__/Query.test.ts
shawnmcknight: I just read your PR description more closely and noted that you specifically call out this potential scenario and that it is documented that a field with 3 potential states needs to be a string with enumerations.

I guess the question mark will be for scenarios where optionality wasn't _really_ intended. That is, it's really a two-state field where `null` and `0` mean the same thing but some logic is relying on basic's interpretation of `null` as false.

